PR #46, opened 2017-06-15
Events endpoint Issue #6

### LINKED ISSUE #6: Create an events endpoint
This will be the first iteration of the endpoint. For now this will return entries from our local database only.

This API will show all events

To do this you should:
- [ ] Create a controller with a `#index` action
- [ ] Create a GET route to `/api/v1/events`
- [ ] Create a `Event` model (with validations)
- [ ] Create a migration

The table should have the follow columns:
* name - string
* description - text
* URL - string
* start_date - datetime
* end_date (optional) - datetime
* address1 - string
* address2 - string
* city - string
* state - string
* zip - string
* scholarship_available - boolean

leenyburger:
# Description of changes
Add event endpoint with index action per issue #6.  This pull request adds an events controller with an index action that renders all events in a json format.   This request also adds a model with validations.  The validation for websites is currently set to start with "www.".  The validation does not catch "www.."  It will reject ftp and other sites not starting with www.  Currently allows URL, end_date, and address2 to be empty.  Also created the events table with a migration.  Added events route to routes.rb.  

# Issue Resolved

Fixes #6

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -9,10 +9,6 @@ class Event < ApplicationRecord
 	VALID_ZIP_REGEX= /\A\d{5}-?\d{4}?\z/
 	validates :zip, presence: true, format:{with:VALID_ZIP_REGEX}
 	validates :scholarship_available, inclusion: {in: [true, false]}
-	VALID_URL_REGEX = /\Awww\.+.+\.[a-z]+\z/i 
-	validates :URL, format: {with:VALID_URL_REGEX}
-
-	
-	
-
+	VALID_URL_REGEX = URI::regexp 
+	validates :url, format: {with:VALID_URL_REGEX}
 end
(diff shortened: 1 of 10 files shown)


app/models/event.rb
```diff
@@ -0,0 +1,18 @@
...
+	validates :description, presence: true
+	validates :start_date, presence: true
+	validates :address1, presence: true
+	validates :city, presence: true
+	validates :state, presence: true, length: {is: 2} #Dropdown selection on front end?  
+	VALID_ZIP_REGEX= /\A\d{5}-?\d{4}?\z/
+	validates :zip, presence: true, format:{with:VALID_ZIP_REGEX}
+	validates :scholarship_available, inclusion: {in: [true, false]}
+	VALID_URL_REGEX = /\Awww\.+.+\.[a-z]+\z/i 
+	validates :URL, format: {with:VALID_URL_REGEX}
```
rickr: What do you think about using `URI::Regexp` here? It would make people include `http://` or `https://` but we would have a tried and true regexp. We could also use `URI.parse` to inspect the input a bit.

PR #52, opened 2017-06-17
Feature/raise error on missing jwt secret - resolves issue 39

### LINKED ISSUE #39: Raise an error if jwt secret is unavailable
# Feature

https://github.com/OperationCode/operationcode_backend/blob/master/app/lib/json_web_token.rb#L8
In order to make debugging easier if the secrets available in the above file are not available we should raise an error with a good error message

Check for the jwt_secret token and raise an error if it doesn't exist.

hpjaj:
# Description of changes
#### Issue resolution
If `jwt_secret` is unavailable, the `rescue` block will capture the error, without halting the program.  	This implementation follows existing [precedence for capturing and logging errors for jwt_secret](https://github.com/OperationCode/operationcode_backend/blob/master/app/lib/json_web_token.rb#L15-L18).

Alternatively, we can be more aggressive, and `raise` an error, if that is the desired approach.  That will, however, halt the program, which might be what you are looking for here?

#### Development/Debugging gem
This PR also adds the Awesome Print gem for development and testing envs.  This speeds up my production when debugging and using the `rails console`, by increasing the readability of any code that is printed out.

It does so by providing structure and highlighting.

Ruby Gems link:  https://rubygems.org/gems/awesome_print

GitHub home page:  https://github.com/awesome-print/awesome_print

# Issue Resolved
Fixes #39

### DIFF
--- a/app/lib/json_web_token.rb
+++ b/app/lib/json_web_token.rb
@@ -6,6 +6,9 @@ class JsonWebToken
     payload['exp'] = expiration.to_i.hours.from_now.to_i
 
     JWT.encode(payload, Rails.application.secrets.jwt_secret)
+  rescue => e
+    Rails.logger.debug "Failed to encode JsonWebToken due to: #{e}"
+    nil
   end
 
   def self.decode(token)
--- a/app/lib/json_web_token.rb
+++ b/app/lib/json_web_token.rb
@@ -7,8 +7,7 @@ class JsonWebToken
 
     JWT.encode(payload, Rails.application.secrets.jwt_secret)
   rescue => e
-    Rails.logger.debug "Failed to encode JsonWebToken due to: #{e}"
-    nil
+    raise "Failed to encode JsonWebToken due to: #{e}"
   end
 
   def self.decode(token)
--- a/Gemfile
+++ b/Gemfile
@@ -33,6 +33,7 @@ group :development, :test do
   gem 'mocha'
   gem 'factory_girl_rails'
   gem 'faker'
+  gem 'awesome_print', '~> 1.8'
 end
 
 group :development do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -46,6 +46,7 @@ GEM
     airtable (0.0.9)
       httparty
     arel (7.1.4)
+    awesome_print (1.8.0)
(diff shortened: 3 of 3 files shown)


hpjaj:
@rickr - CI/travis-ci/push is failing.  The issue appears to be in [lines 514 - 17](https://travis-ci.org/OperationCode/operationcode_backend/builds/244085422?utm_source=github_status&utm_medium=notification):

```
rake aborted!
PG::ConnectionBad: could not connect to server: Connection refused
	Is the server running on host "operationcode-psql" (172.18.0.3) and accepting
	TCP/IP connections on port 5432?
```

I assume this is this Docker related?  Recommendations on resolving this?

rickr:
@hpjaj - Raising an error here is indeed what we want. If this secret doesn't exist an exception is raised elsewhere (IIRC in the jwt gem) and the cause is not so obvious. The intent of this ticket is to catch that sooner and raise a more clear error message.

app/lib/json_web_token.rb
```diff
@@ -6,6 +6,8 @@ def self.encode(payload, expiration = nil)
     payload['exp'] = expiration.to_i.hours.from_now.to_i
 
     JWT.encode(payload, Rails.application.secrets.jwt_secret)
+  rescue => e
+    raise "Failed to encode JsonWebToken due to: #{e}"
```
hpjaj: @rickr - Is this more inline with what you were looking for?  This will raise if `jwt_secret` is missing, as well as any other issues in the `encode` method.

Or did you want to `raise` **only if** `jwt_secret` is `nil`?  i.e.

```ruby
  def self.encode(payload, expiration = nil)
    expiration ||= Rails.application.secrets.jwt_expiration_hours

    payload = payload.dup
    payload['exp'] = expiration.to_i.hours.from_now.to_i

    raise "Rails.application.secrets.jwt_secret is nil. Discovered in JsonWebToken.encode." if Rails.application.secrets.jwt_secret.nil?

    JWT.encode(payload, Rails.application.secrets.jwt_secret)
  end
```

rickr: Yea - I think this version will help a ton! Thanks for the quick turn around!

PR #68, opened 2017-06-20
Update code_schools.yml

kylemh:
# Description of changes
Remove old entry for Code Platoon and adjust PDX Code Guilds to VA-Approved status.

# Issue Resolved
Fixes https://github.com/OperationCode/operationcode_frontend/issues/180

### DIFF
--- a/config/code_schools.yml
+++ b/config/code_schools.yml
@@ -329,20 +329,6 @@
       state: CA
       zip: 95113
 
-- name: Code Platoon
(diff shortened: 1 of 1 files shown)


kylemh:
Sorry for the weird commit message on 943484d - I didn't want there to be an issue #180 in the future that this some how messes up. I don't know if that's even possible, but it definitely won't be now.

PR #70, opened 2017-06-22
Feature/id me verification endpoint

### LINKED ISSUE #38: Create ID.me verification endpoint
# Feature

We need to support verifying users via ID.me

Create an endpoint `/api/v1/users/profile/verify`
You will be passed one parameter `access_token`.
Take that parameter and GET the following URL to `https://api.id.me/api/public/v2/attributes.json ?access_token=ACCESS_TOKEN`
That should provide you with the following payload:
```
{
  "id": "a1b2c3d4e5f6g7h8i9j0",
  "verified": true,
  "affiliation": "Service Member",
  "fname": Test,
  "lname": User,
  "email": "test.user@id.me",
  "uuid": "a1b2c3d4e5f6g7h8i9j0"
}
```
The data from `verified` should be stored as an attribute on the `User` determined via the JWT

hpjaj:
# Description of changes
Creates an ID.me verification endpoint.

- Creates an API endpoint for the frontend to hit, passing an `access_token`
- Creates an `IdMe` class that works with the ID.me API
- On the `IdMe` class, creates a method to `GET` their verification endpoint
- Updates the `current_user` with ID.me's response
- Updates the frontend with the appropriate response

# Issue Resolved
Fixes #38

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,7 @@ gem 'rails', '~> 5.0.2'
 gem 'redis', '~> 3.0'
 gem 'sentry-raven'
 gem 'sidekiq'
+gem 'httparty'
 
 
 group :development, :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -197,6 +197,7 @@ DEPENDENCIES
   faker
   geocoder
   gibbon
+  httparty
   jwt
   listen (~> 3.0.5)
   lograge
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -15,10 +15,15 @@ module Api
 
       def verify
         verified = IdMe.verify! params[:access_token]
+
+        # FIXME use current user
+        #current_user.update! verified: verified
+
+        User.last.update! verified: verified
         current_user.update! verified: verified
         render json: { status: :ok, verified: verified }
       rescue => e
-        Rails.logger.debug "When verifying User id #{current_user.id} through ID.me, experienced this error: #{e}"
+        Rails.logger.debug "When verifying User id #{User.last.id} through ID.me, experienced this error: #{e}"
         render json: { status: :unprocessable_entity }
       end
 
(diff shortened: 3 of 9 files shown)


alexspence:
Apparently private by itself changes the scope - code climate said it was useless.. @rickr I'll let you merge this in case you want to revert my changes.

hpjaj:
@rickr @alexspence - IMO, Code Climate is a fantastic tool to help ensure quality, consistent code across teams, but it is not gospel.  This is a case in point where code climate, IMO, is completely incorrect.

Ruby classes, and Rails conventions, customarily leverage `private` to delineate between methods that are purposefully part of the **public** API, and private methods that are used to support those public facing methods, but should not be available as public methods themselves.

We should explore how to make adjustments to the code climate rules being used, so that if OC decides to remove a rule, it can be done.

In this case, I feel we should reinstate `private` in both of these controllers, and for expediency, we can set this as a `WONTFIX`.

alexspence:
I agree with you hpjaj, originally I thought it was just a typo, do you want to go ahead and commit the change?

rickr:
We can add it back to the controller but the one in `app/lib/id_me.rb` was the superfluous one. To make a class method private you need to either use `private_class_method`. I think opening self and using `private` works as well.

hpjaj:
Hey @rickr !  How else can I help with this PR?

I believe I read somewhere in Slack that this was being tested in `staging`.  I can help with smoke testing this PR if you can outline the OC process for:

- deploying a PR to staging
- logging into staging (i.e. SSH into something?, URL to visit, etc.)
- logging into the staging `rails console` and command line

For common ground, here is what I am accustomed to doing at work for the above:

- Push current code to the remote PR on GitHub
- Connect to our work VPN
- Then run a `capistrano` command to deploy the PR to the staging env
- Once the deploy is complete, browse to the staging env's URL and login
- On the command line, SSH into the staging environment
- Enter credentials to access the env
- `cd` into `current` directory
- Login to the staging `rails console`, or `tail` the log, etc.

From there I'll need to test hitting the ID.me API.  I was never able to do so locally.  If you have been successful doing so from staging, can you also please include instructions on how to accomplish that?

If you can provide me with the equivalent steps, I can help with smoke testing this PR, and refactoring the code accordingly.  

Once you've relayed the above, I'm happy to document all of the above in our docs (our `README`, or wherever is most appropriate). Thanks!

PR #75, opened 2017-07-02
Add dredd for continuous API tests

hollomancer:
# Description of changes
Adds dredd, which will run API tests against the build, created from our documentation in our API blueprint (apiary.apib)

We may want to throw this in Docker instead, I'm not familiar enough with the backend.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,9 @@ language: c
 
 services:
   - docker
+  
+before_install:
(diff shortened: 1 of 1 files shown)


rickr:
This build failed with the following errors:

```
Error: Must specify path to API description document.
Error: Must specify URL of the tested API instance.
```

hollomancer:
later.

PR #86, opened 2017-07-14
Adds send grid job. Fixes #49

### LINKED ISSUE #49: Create an 'add_to_sendgrid' job
# Feature

To support operationcode/operationcode#751 we will need to create a new job that is called when a user joins. It should hit the following

Add a new ActiveJob that adds a user to the sendgrid contactdb. The API endpoint you need is documented [here](https://sendgrid.com/docs/API_Reference/Web_API_v3/Marketing_Campaigns/contactdb.html#Add-Single-Recipient-POST).

Update the `welcome_user` method in the `User` model to call this new job.

There is a sendgrid gem that could be useful: [sendgrid-ruby](https://github.com/sendgrid/sendgrid-ruby#quick_start)

rickr:
# Description of changes

Adds a new job to run when a user signs up to add them to our sendgrid contact list.

# Issue Resolved

Fixes #49

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,6 +23,7 @@ gem 'pundit'
 gem 'rack-cors'
 gem 'rails', '~> 5.0.2'
 gem 'redis', '~> 3.0'
+gem 'sendgrid-ruby'
 gem 'sentry-raven'
 gem 'sidekiq'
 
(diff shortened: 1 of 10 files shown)


rickr:
@hpjaj Fixes are in.

PR #92, opened 2017-07-18
feature/51-send-email-with-sendgrid-template

### LINKED ISSUE #51: Convert to sendgrid API instead of SMTP
# Feature

We have white labeled IP addresses and can now use sendgrid's API to track our emails.

~Integrate and convert this app to use [sendgrid-actionmailer](https://github.com/eddiezane/sendgrid-actionmailer)~
I think it'll be easier/simpler to add on to the new SendGridClient located in `app/lib/send_grid_client.rb`.

The transactional template id is `e6bc900f-fc72-4891-8f7a-8f3f3c9c105b`. It takes one substitution: `%user_count%` - which lists the number of users we have.

This ticket should create a new method for the SendGridClient that sends a transactional email using [sendgrid-ruby](https://github.com/sendgrid/sendgrid-ruby/tree/master/examples). This method should take a `transactional_template_id` and a hash of values that will be substituted into the template.

Once deployed we can test to make sure emails are delivered as expected.

Another ticket will be filed to change over delivery from SMTP to this new method.

hpjaj:
# Description of changes
Creates a new method for the SendGridClient that sends a transactional email using sendgrid-ruby. This method takes a transactional_template_id and a hash of values that can be substituted into the template.

# Issue Resolved
Fixes #51

### DIFF
--- a/app/lib/send_grid_client.rb
+++ b/app/lib/send_grid_client.rb
@@ -17,6 +17,52 @@ class SendGridClient
     Rails.logger.error e.backtrace.join("\n")
   end
 
+  # Sends a transactional email with a template, based on the passed in options hash.
+  #
+  # @param options [Hash] Sample options hash:
+  #   {
+  #     to: 'tom@gmail.com',
+  #     from: 'frank@oc.com',
+  #     subject: 'Hello',
+  #     user_count: 5,
+  #     body: 'Hello world!',
+  #     template_id: 'asdfas-123123-asdfasf-2qwerqwe'
+  #   }
+  # @return [SendGrid::Response] Includes a body and status_code
+  # @see https://github.com/sendgrid/sendgrid-ruby/blob/7c8973f7b8de51a98dce73a52138791b19434573/USE_CASES.md#with-mail-helper-class
+  #
+  def send_email_with_template(options={})
+    Rails.logger.info("Sending email to #{options[:to]} at #{Time.current} re: #{options[:subject]}")
+
+    mail = {
+      "personalizations" => [
+        {
+          "to" => [
+            {
+              "email" => options[:to]
+            }
+          ],
+          "subject" => options[:subject],
+          "template_id" => options[:transactional_template_id]
+        }
+      ],
+      "from" => {
+        "email" => options[:from]
+      },
+      "content" => [
+        {
+          "type" => "text/plain",
+          "value" => options[:body]
+        }
+      ]
+    }
+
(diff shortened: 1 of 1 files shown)


hpjaj:
@rickr - I was having issues with the substitutions portion during testing:

"substitutions" => {
  "-user_count-" => options[:user_count]
}

We will have to investigate this further.

app/lib/send_grid_client.rb
```diff
@@ -17,6 +17,52 @@ def add_user(user)
...
+          "type" => "text/plain",
+          "value" => options[:body]
+        }
+      ]
+    }
+
+    @sendgrid.client.mail._("send").post(request_body: mail)
+  rescue => e
+    Rails.logger.error "Failed to send email to #{options[:to]} re: #{options[:subject]} due to: #{e}"
+    Rails.logger.error e.backtrace.join("\n")
```
rickr: Can we reraise the error here so we know if it fails?

hpjaj: @rickr - Done https://github.com/OperationCode/operationcode_backend/pull/92/commits/419f7f8bbe55d07c572bbbbd512cc3294099b515

hpjaj: Now we just need a solution for this:

> @rickr - I was having issues with the substitutions portion during testing:
> 
> "substitutions" => {
> "-user_count-" => options[:user_count]
> }
> 
> We will have to investigate this further.

hpjaj: Resolved the `user_count` issue: https://github.com/OperationCode/operationcode_backend/pull/92/commits/6d5668d581e73e602f189b55dc54ad2fa3771f84

hpjaj:
@rickr - this one is ready

app/lib/send_grid_client.rb
```diff
@@ -43,6 +43,9 @@ def send_email_with_template(options={})
             }
           ],
           "subject" => options[:subject],
+          "substitutions" => {
+            "-user_count-" => options[:user_count]
```
rickr: Since this requires a string can you add a `.to_s` to this line to make sure we always pass a string to the API?

hpjaj:
@rickr - this one is ready

app/lib/send_grid_client.rb
hpjaj: Done: https://github.com/OperationCode/operationcode_backend/pull/92/commits/4bf02cb0bdd2058daa4e42a14477ef404f304eb5

PR #101, opened 2017-07-20
SSO Redirect

### LINKED ISSUE #91: Implement SSO for discourse
# Feature

In order to provide a seamless UX between oc.org and discourse we need to implement SSO for discourse.

We should send all info needed for SSO discourse into the JWT.

These links should help:
https://meta.discourse.org/t/official-single-sign-on-for-discourse/13045
https://stackoverflow.com/questions/25478510/incorporating-discourse-sso-with-existing-rails-site-with-devise

rickr:
# Description of changes

Adds logic for discourse SSO logins

# Issue Resolved

Fixes #91

### DIFF
--- a/app/controllers/api/v1/sessions_controller.rb
+++ b/app/controllers/api/v1/sessions_controller.rb
@@ -5,17 +5,43 @@ module Api
       before_action :set_default_response_format
 
       def create
-        Rails.logger.debug 'In Create'
         self.resource = warden.authenticate!(auth_options)
         sign_in(resource_name, resource)
-        render json: { token: resource.token,
-                       user: UserSerializer.new(current_user) }
+
+        set_sso_response if sso_request?
+        @redirect_path ||= '/profile'
+
+        render json: {
+          token: resource.token,
+          user: UserSerializer.new(current_user),
+          redirect_to: @redirect_path
+        }
       end
 
       def set_default_response_format
         request.format = 'json'
       end
 
+      private
+
+      def sso_request?
+        return true if params[:sso].present? && params[:sig].present?
+      end
+
+      def sso_params
+        { sso: params[:sso], sig: params[:sig] }
+      end
+
+      def set_sso_response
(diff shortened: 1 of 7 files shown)


app/controllers/api/v1/sessions_controller.rb
```diff
@@ -5,17 +5,43 @@ class SessionsController < Devise::SessionsController
...
       end
 
       def set_default_response_format
         request.format = 'json'
       end
 
+      private
+
+      def sso_request?
+        return true if params[:sso].present? && params[:sig].present?
```
hpjaj: Perhaps remove the `return true if` to return either `boolean` case:

```ruby
def sso_request?
  params[:sso].present? && params[:sig].present?
end

File: app/lib/discourse/single_sign_on.rb [hunk 2]
```diff
@@ -0,0 +1,108 @@
...
+
+    def self.sso_secret
+      raise RuntimeError, "sso_secret not implemented on class, be sure to set it on instance"
+    end
+
+    def self.sso_url
+      raise RuntimeError, "sso_url not implemented on class, be sure to set it on instance"
+    end
+
+    def self.parse(payload, sso_secret = nil)
```
hpjaj: This method has a few chunks of logic.  Can you break some of it out into smaller, single purpose `private` methods?

app/controllers/api/v1/sessions_controller.rb
rickr: 7ec989a

app/lib/discourse/single_sign_on.rb
rickr: I don't love this code either but it's the reference implementation from discourse: https://github.com/discourse/discourse/blob/master/lib/single_sign_on.rb

I opted out of making any changes (except for the necessary ones for our environment) to make any troubleshooting we might need to do easier. I dropped a comment here noting this: https://github.com/OperationCode/operationcode_backend/blob/3b98c8371a351a2b20f3727ebc9bfbcc4b7e6001/app/lib/discourse/single_sign_on.rb#L4

PR #108, opened 2017-07-21
Feature/105 save users state

### LINKED ISSUE #105: Save user's state
# Feature

When a user signs up they give their zip. We need to be able to query the number of users in a state.

Add a new attribute to the `User` model called `state`.
Extend the `before_save` filter to save the user's state into the new attribute.
It should be stored in the 2 letter format

This should also include a rake task to populate all existing users adding states if their latitude and longitude exists as we have some users that don't have zips associated with them.

There is a rate limit on the geocoding API so there should be a sleep or other kind of throttling:
https://developers.google.com/maps/documentation/geocoding/usage-limits

hpjaj:
# Description of changes
Saves the state the `User` lives in, to their db record.  Specifically:

- Adds a new attribute to the `User` model called `:state`
- Extends the `before_save` filter to save the user's `:state` abbreviation into the new attribute
- Rake task to populate all existing users `:state`, if their latitude and longitude exists
- Rake task accounts for Google geocoding API  rate limits 

# Issue Resolved
Fixes #105

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -5,7 +5,9 @@ class User < ApplicationRecord
          :recoverable, :rememberable, :trackable, :validatable
 
   geocoded_by :zip do |user, results|
-    if geocoded_object = results.first
+    geocoded_object = results.first
+
+    if geocoded_object.present?
       user.latitude  = geocoded_object.latitude
       user.longitude = geocoded_object.longitude
       user.state     = geocoded_object.state_code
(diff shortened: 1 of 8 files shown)


app/models/user.rb
```diff
@@ -3,7 +3,16 @@ class User < ApplicationRecord
   # :confirmable, :lockable, :timeoutable and :omniauthable
   devise :database_authenticatable, :registerable,
          :recoverable, :rememberable, :trackable, :validatable
-  geocoded_by :zip
+
+  geocoded_by :zip do |user, results|
```
hpjaj: @rickr - Not terribly pleased that this is a block being passed, at the top of the `user.rb` file, versus a method being called.  However, was not able to find an alternative solution in the [Geocoder docs](https://github.com/alexreisner/geocoder).  

Let me know if you know of one/want it changed.

PR #109, opened 2017-07-21
Feature/104 user count by state

### LINKED ISSUE #104: Add state param to users/by_location
# Feature

We should be able to get a count of users in a specific state. Right now it looks like all searches are done by a radius from a point.

Add a new param to the `users/by_location` action called `state`. It should perform a query on the users table and return a count of the users in a particular state.
Make sure to update teh docs

**This is blocked by #105**

hpjaj:
# Description of changes
Adds ability to query by users' `:state`'s, to determine user count.  Specifically:

- Adds a new `param` to the `users/by_location` action called `:state` that returns a count of the users in a particular state.
- Updates the API docs 
- Additional test coverage

# Issue Resolved
Fixes #104

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,10 @@ db_migrate:
 db_rollback:
 	docker-compose run ${RAILS_CONTAINER} rake db:rollback
 
+.PHONY: db_seed
+db_seed:
+	docker-compose run ${RAILS_CONTAINER} rake db:seed
+
 .PHONY: test
 test: bg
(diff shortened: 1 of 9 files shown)


hpjaj:
@rickr - Travis is failing; however, when grepping for `error` or `failure`, there are no failures or errors associated with any tests.

Can you determine if something is going on with the build, or?

hpjaj:
@rickr - this is happening again:

https://github.com/OperationCode/operationcode_backend/pull/109#issuecomment-317047724

Can you look into it?

rickr:
I don't see any problems either. Are we mocking the call to the geocoding API? I wonder if we're getting rate limited?

hpjaj:
@rickr - this one is ready

PR #124, opened 2017-08-02
Create scholarship models#122

### LINKED ISSUE #122: Create scholarship models
# Feature

See operationcode/operationcode_frontend#357 for some more background.

Add a new scholarship model.
It should follow the pattern of this form: https://airtable.com/shroOCkoOu7vOXuyN

Scholarships should have the following attributes:
name - string
description - text
location - string
terms - text
start_date - datetime
end_date - datetime

Scholarships should have many applications
Users should have many applications

Applications should have the following attributes:
name (delegated to user)
email (delegated to user)
reason - text
terms_accepted - boolean

Only verified members should be able to submit an application.

jmayergit:
# Description of changes
Generates Scholarship Model
Generates ScholarshipApplication Model
Adds delegation to ScholarshipApplication
Removes unused test files

# Issue Resolved

Fixes #122

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -1,4 +1,6 @@
 class Application < ApplicationRecord
+  belongs_to :user
+  belongs_to :scholarship
   # before_save only verified members can submit
   delegate :name, to: :user
   delegate :email, to: :user
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -1,3 +1,5 @@
 class Application < ApplicationRecord
   # before_save only verified members can submit
+  delegate :name, to: :user
+  delegate :email, to: :user
 end
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -1,0 +1,3 @@
+class Application < ApplicationRecord
+  # before_save only verified members can submit
+end
--- a/app/models/scholarship_application.rb
+++ b/app/models/scholarship_application.rb
@@ -1,7 +1,15 @@
 class ScholarshipApplication < ApplicationRecord
+  validate :user_is_verified
+
   belongs_to :user
   belongs_to :scholarship
-  # before_save only verified members can submit
+
   delegate :name, to: :user
   delegate :email, to: :user
+
+  def user_is_verified
+    unless self.user.verified
+      errors.add(:verified, "Only verified users may submit applications")
+    end
+  end
 end
--- a/app/models/scholarship_application.rb
+++ b/app/models/scholarship_application.rb
@@ -1,4 +1,4 @@
-class Application < ApplicationRecord
+class ScholarshipApplication < ApplicationRecord
   belongs_to :user
   belongs_to :scholarship
   # before_save only verified members can submit
--- a/app/models/scholarship_application.rb
+++ b/app/models/scholarship_application.rb
@@ -4,11 +4,12 @@ class ScholarshipApplication < ApplicationRecord
   belongs_to :user
   belongs_to :scholarship
 
-  delegate :name, to: :user
-  delegate :email, to: :user
+  delegate :name, to: :user, prefix: true
+  delegate :email, to: :user, prefix: true
+  delegate :verified, to: :user, prefix: true
 
   def user_is_verified
-    unless self.user.verified
+    unless self.user_verified
       errors.add(:verified, "Only verified users may submit applications")
     end
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -76,6 +76,10 @@ class User < ApplicationRecord
     "#{first_name} #{last_name}"
   end
 
+  def email
+    self[:email]
+  end
+
   def welcome_user
     invite_to_slack
     add_to_airtables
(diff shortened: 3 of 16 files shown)


app/models/application.rb
```diff
@@ -0,0 +1,7 @@
+class Application < ApplicationRecord
```
rickr: I think we should probably rename this to `ScholarshipApplication` or something like that. `Application` might be confused with the actual rails application.

app/models/user.rb
```diff
@@ -75,6 +76,10 @@ def name
     "#{first_name} #{last_name}"
   end
 
+  def email
```
rickr: This method isn't needed - you should be able to access the email attribute without it.

app/models/application.rb
```diff
@@ -0,0 +1,7 @@
...
+  belongs_to :user
+  belongs_to :scholarship
+  # before_save only verified members can submit
```
rickr: This would end up being a validation on the user's validated attribute.

You can read about rails validations here: http://edgeguides.rubyonrails.org/active_record_validations.html#validation-helpers

db/migrate/20170802201510_create_scholarships.rb
```diff
@@ -0,0 +1,14 @@
+class CreateScholarships < ActiveRecord::Migration[5.0]
+  def change
+    create_table :scholarships do |t|
+      t.string :name
+      t.text :description
+      t.string :location
+      t.text :terms
+      t.datetime :start_date
```
hpjaj: If we want `start_date` and `end_date` to truly be dates, we should change these attributes to be of type `t.date`.  

If not, and we want them to be a `datetime`, perhaps rename the attributes to be more telling of the data they store?

app/models/application.rb
jmayergit: That's funny I actually got confused myself on that very issue

db/migrate/20170802201510_create_scholarships.rb
jmayergit: I'll change them to type date

rickr: I think date time is better since we can say something like 'Applications close at 5pm 8/2'. Maybe a better name is in order? open_time and close_time?

jmayergit: Okay

test/factories/scholarships.rb
```diff
@@ -0,0 +1,10 @@
+FactoryGirl.define do
+  factory :scholarship do
+    name "MyString"
+    description "MyText"
+    location "MyString"
+    terms "MyText"
+    start_date "2017-08-02 15:15:11"
```
hpjaj: Sneaky factories...can you rename the two columns here, too.

test/models/scholarship_application_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
+
+class ScholarshipApplicationTest < ActiveSupport::TestCase
```
hpjaj: This is user preference, whether or not to keep these two autogenerated `*_test.rb` files or not.  

I typically delete them, until I have tests in them.  Some prefer to keep them as placeholders.

app/models/scholarship_application.rb
```diff
@@ -0,0 +1,15 @@
...
+  validate :user_is_verified
+
+  belongs_to :user
+  belongs_to :scholarship
+
+  delegate :name, to: :user
+  delegate :email, to: :user
+
+  def user_is_verified
+    unless self.user.verified
```
hpjaj: A style option, nothing major.  For these `boolean` attributes, you could also end them with a `?`, and it will do the same thing; however, it sort of grammatically appears as if you are asking a yes/no question.   For example:

```ruby
  unless self.user.verified?
```

app/models/scholarship_application.rb
```diff
@@ -0,0 +1,15 @@
+class ScholarshipApplication < ApplicationRecord
+  validate :user_is_verified
+
+  belongs_to :user
+  belongs_to :scholarship
+
+  delegate :name, to: :user
```
hpjaj: Perhaps add the `prefix: true` option to these two delegate methods, to be more explicit about these two methods' functionality?  

For example:
```ruby
delegate :name, to: :user, prefix: true
```

That way you would call `scholarship_application.user_name` instead of `scholarship_application.name`.

test/factories/scholarships.rb
jmayergit: woops, Yep I will, renaming stuff is a bit of a nightmare

test/models/scholarship_application_test.rb
jmayergit: There's one other empty test stub, service_test.rb, I feel like I should remove that one if I remove these two

hpjaj: ➕ 1

app/models/scholarship_application.rb
rickr: Since this is an attribute on the `User` model this is a good candidate for delegation as well.

That would let us do `scholarship_application.user_verified`

jmayergit: Interestingly enough delegation throws an error when using `?`, so I'll leave it out.

jmayergit:
Are all the necessary changes in place? @rickr

PR #131, opened 2017-08-06
Add staff / volunteers table

### LINKED ISSUE #130: Add team members to db seeds
# Bug Report
When running the backend locally, the staff are not being rendered on the team page.

Staff / Volunteers are displayed

sethbergman:
Added table
# Issue Resolved

Fixes #130

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -130,6 +130,13 @@ ActiveRecord::Schema.define(version: 20170723213338) do
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
   end
+  
+  create_table "staff", force: :cascade do |t|
+    t.string   "name"
+    t.string   "role"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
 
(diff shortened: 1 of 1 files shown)


db/schema.rb
```diff
@@ -130,6 +130,13 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
   end
+  
+  create_table "staff", force: :cascade do |t|
```
hpjaj: Unfortunately you cannot update `schema.rb` directly in order to add/update db tables.  Instead you'll need to:

1. Create a migration file that creates the new table.  Here's [an example](https://github.com/OperationCode/operationcode_backend/blob/master/db/migrate/20170701172700_create_resources.rb).
2. Run the migration task to update the `schema` file.  With docker running, you can call the [make db_migrate](https://github.com/OperationCode/operationcode_backend/blob/master/Makefile#L35) command.

sethbergman: I'm going to hand this over to you guys to see how you want to handle it.

rickr:
This table already exists.

PR #136, opened 2017-08-12
added DevPoint Labs to code school yml file

nightwing891:
# Description of changes
add DevPoint Labs to the VA approved section of the site. Also if we can use this logo attach below on it, that would be great.

Thanks

dpl_logo


kylemh:
I've approved this, but I'm not certain if the changes need to manually updated against our database, so I'll leave it to @rickr for the merge.

PR #145, opened 2017-08-18
Create scholarship controller

### LINKED ISSUE #144: Add Scholarship[&Applications] controllers/routes
# Feature
Adds Scholarship and ScholarshipApplications endpoints to API
Backend support for the scholarship application flow

Accept incoming GET requests for /scholarships & /scholarships/:id
Accept incoming POST requests for /scholarship_applications

jmayergit:
# Description of changes

Adds Scholarship and ScholarshipApplications endpoints to API so that verified users can apply for scholarships

# Issue Resolved

Fixes #144

### DIFF
--- a/app/controllers/api/v1/scholarship_applications_controller.rb
+++ b/app/controllers/api/v1/scholarship_applications_controller.rb
@@ -8,7 +8,7 @@ module Api
         if @scholarship_application.save
           render json: { redirect_to: '/success' }
         else
-          render json: @scholarship_application.errors, status: :unauthorized
+          render json: { errors: @scholarship_application.errors.values }, status: :bad_request
         end
       end
     end
--- a/app/controllers/api/v1/scholarship_applications_controller.rb
+++ b/app/controllers/api/v1/scholarship_applications_controller.rb
@@ -1,0 +1,16 @@
+module Api
+  module V1
+    class ScholarshipApplicationsController < ApplicationController
+      respond_to :json
+
+      def create
+        @scholarship_application = ScholarshipApplication.new(reason: params[:reason], terms_accpeted: params[:terms_accepted], scholarship_id: params[:scholarship_id], user_id: current_user.id)
+        if @scholarship_application.save
+          render json: { redirect_to: '/success' }
+        else
+          render json: @scholarship_application.errors, status: :unauthorized
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/api/v1/scholarship_applications_controller.rb
+++ b/app/controllers/api/v1/scholarship_applications_controller.rb
@@ -1,15 +1,18 @@
 module Api
   module V1
     class ScholarshipApplicationsController < ApplicationController
-      respond_to :json
-
       def create
-        @scholarship_application = ScholarshipApplication.new(reason: params[:reason], terms_accpeted: params[:terms_accepted], scholarship_id: params[:scholarship_id], user_id: current_user.id)
-        if @scholarship_application.save
-          render json: { redirect_to: '/success' }
-        else
-          render json: { errors: @scholarship_application.errors.values }, status: :bad_request
-        end
+        current_user.scholarship_applications.create! scholarship_application_params
+
+        render json: { redirect_to: '/success' }, status: :created
+      rescue StandardError => e
+        render json: { error: e.message }, status: :bad_request
+      end
+
+      private
+
+      def scholarship_application_params
+        params.require(:scholarship_application).permit(:reason, :terms_accepted, :scholarship_id)
       end
     end
   end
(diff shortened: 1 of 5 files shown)


app/controllers/api/v1/scholarship_applications_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+  module V1
+    class ScholarshipApplicationsController < ApplicationController
+      respond_to :json
+
+      def create
+        @scholarship_application = ScholarshipApplication.new(reason: params[:reason], terms_accpeted: params[:terms_accepted], scholarship_id: params[:scholarship_id], user_id: current_user.id)
+        if @scholarship_application.save
+          render json: { redirect_to: '/success' }
+        else
+          render json: { errors: @scholarship_application.errors.values }, status: :bad_request
```
hpjaj: Since this `@scholarship_application` instance variable is only being used locally, and does not need to be accessible in a view, it can just be a local variable, i.e. `scholarship_application`.

app/controllers/api/v1/scholarship_applications_controller.rb
```diff
@@ -0,0 +1,16 @@
+module Api
+  module V1
+    class ScholarshipApplicationsController < ApplicationController
+      respond_to :json
+
+      def create
+        @scholarship_application = ScholarshipApplication.new(reason: params[:reason], terms_accpeted: params[:terms_accepted], scholarship_id: params[:scholarship_id], user_id: current_user.id)
```
hpjaj: Instead of passing in all of the attributes here, you can leverage strong params.  For example:

```ruby
def scholarship_application_params
  params.require(:scholarship_application).permit(:reason, :terms_accepted, :scholarship_id, :user_id)
end
```

If the params do not have the attributes nested within a `scholarship_application` key (i.e. 
`{ scholarship_application: { reason: ..., user_id: ... } }`, then you can do a less strict version of the strong params like this:

```ruby
def scholarship_application_params
  params.permit(:reason, :terms_accepted, :scholarship_id, :user_id)
end
```


And then you can just call:

```ruby
ScholarshipApplication.create! scholarship_application_params
```

app/controllers/api/v1/scholarship_applications_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+        if @scholarship_application.save
+          render json: { redirect_to: '/success' }
+        else
```
hpjaj: Instead of the conditional (`else`) here, you can use the `create!` with a `!`, and then `rescue` any errors, and `render` them to the frontend.

Here is an example of that in the [`api/v1/team_members_controller`](https://github.com/OperationCode/operationcode_backend/blob/master/app/controllers/api/v1/team_members_controller.rb#L10-L16).

hpjaj:
There is a typo in one of the attributes in the `ScholarshipApplication` db table: [`terms_accpeted`](https://github.com/OperationCode/operationcode_backend/blob/master/db/schema.rb#L64).

app/controllers/api/v1/scholarships_controller.rb
```diff
@@ -0,0 +1,13 @@
+module Api
+  module V1
+    class ScholarshipsController < ApplicationController
+      def index
+        render json: Scholarship.all
+      end
+
+      def show
+        render json: Scholarship.find(params[:id])
```
hpjaj: The `find` method can error out, if it does not find the record, for whatever reason.  To avoid that, one solution would be to use the `find_by` method.  i.e. `Scholarship.find_by(id: params[:id])`.  

If it does not find a record that meets the criteria, it just returns `nil`.  Give these a try in the rails console.

That said, you might want to pair this with some error protection (i.e. `if it finds the record, render it.  else, render some error sentence`).

app/controllers/api/v1/scholarship_applications_controller.rb
jmayergit: Awesome, I really appreciate it Harry

jmayergit: Since I don't send `:user_id` from the front, in other words since we don't receive it as a paramter, I went ahead and did this

```ruby
params.require(:scholarship_application).permit(:reason, :terms_accepted, :scholarship_id).merge(user_id: current_user.id)
```

app/controllers/api/v1/scholarships_controller.rb
jmayergit: Okie dokie

app/controllers/api/v1/scholarship_applications_controller.rb
hpjaj: Creative, you could do that.  It feels like a little misdirection IMO.  

What about building the `ScholarshipApplication` object directly off the `current_user`?  Like this:

```ruby
current_user.scholarship_applications.create! scholarship_application_params
```

jmayergit: Nice!

PR #147, opened 2017-08-21
Fix user count

hollomancer:
Currently, the email displays '0+' users. I just jumped in and made it '3000+' as per our current database.

### DIFF
--- a/app/views/user_mailer/welcome.html.erb
+++ b/app/views/user_mailer/welcome.html.erb
@@ -11,7 +11,7 @@
 </p>
 
 <p>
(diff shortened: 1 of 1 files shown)


app/views/user_mailer/welcome.html.erb
```diff
@@ -11,7 +11,7 @@
 </p>
 
 <p>
-  The Operation Code website uses Ruby On Rails. If you want to dive in and help, you can get started by reading our <a href="https://github.com/OperationCode/operationcode/blob/master/CONTRIBUTING.md">docs</a>, or asking around in the #operationcode-dot-org Slack channel.
+  The Operation Code website uses React, a Javascript framework, and Ruby On Rails. If you want to dive in and help, you can get started by reading our <a href="https://github.com/OperationCode/operationcode/blob/master/CONTRIBUTING.md">docs</a>, or asking around in the #operationcode-dot-org Slack channel.
```
Exupery: Just a minor/nitpicky suggestion: JavaScript is the proper styling (not Javascript) https://developer.mozilla.org/en-US/docs/Web/JavaScript

Issue #118, opened 2017-07-24
Emails should be downcased before saving


rickr:
# Feature
## Why is this feature being added?

Emails should be stored in all lowercase in the database.

## What should your feature do?
Add a `before_save` filter to downcase email addresses on the `User` model.

sabrown84:
Hi I'd like to try this one out. Is there any additional information that you think I will need? I noticed that in the user model has two validations for emails should I look to add a validation for this lowercase feature as well?

ianroberts131:
I've actually been trying to get this setup, but having some issues getting my tests to run. Looks like @sabrown84 is going to take this 1, but I'd love to help on other issues so would still like to resolve the problems I'm having with the tests.

When I run `rake test`, I get the following error:

```
rake test
rake aborted!
Bundler::GemNotFound: Could not find nokogiri-1.7.1 in any of the sources
```

When I `bundle install`, I then get the following:

```
Gem::InstallError: nokogiri requires Ruby version < 2.4, >= 2.1.0.
An error occurred while installing nokogiri (1.7.1), and Bundler cannot
continue.
Make sure that `gem install nokogiri -v '1.7.1'` succeeds before bundling.
```

Any ideas? Thanks for your help.

sabrown84:
Ok @ianroberts131 I haven't started setting it up I was just starting to think it through before jumping in. I may run into the same error...

hpjaj:
@ianroberts131 - You can run the tests through [Docker](https://www.docker.com/).  Try this:

1. Start Docker
2. run `make setup`
3. run `make test`

ianroberts131:
Thanks @hpjaj, that got me going in the right direction. Unfortunately, the version of Make installed [here](http://gnuwin32.sourceforge.net/packages/make.htm), per the `CONTRIBUTING.MD` file, is one version before a bug I was facing got fixed. Here's the error message I was getting:

```
C:\Program: Interrupt/Exception caught (code = 0xc00000fd, addr = 0x4227d3)
```

Ultimately was able to get it working by running make via MinGW, like so:

```
mingw32-make.exe test
```

Anyway, this isn't really relevant since @sabrown84 is taking this issue, but figured I would report back in case she or anyone else faces the same issue.

sabrown84:
Thank you @ianroberts131 and @hpjaj. I thought I would have a similar problem, but I didn't.

hpjaj:
[...]

PR #148, opened 2017-08-22
Feature/118 emails downcased

sabrown84:
# Description of changes
 Emails should be stored in all lowercase into the database 

# Issue Resolved
 Issue #118 Emails should be downcased using `before_save` within the `User` model and tested within the`user_test.rb`

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -20,6 +20,7 @@ class User < ApplicationRecord
   after_create :welcome_user
   before_save :geocode, if: ->(v) { v.zip.present? && v.zip_changed? }
   before_save :upcase_state
+  before_save :downcase_email
 
   validates_format_of :email, :with => /\A.*@.*\z/
   validates :email, uniqueness: true
@@ -102,5 +103,9 @@ class User < ApplicationRecord
 
   def upcase_state
    state.upcase! if state
+  end
+
+  def downcase_email
+    email.downcase! if email
   end
 end
--- a/test/models/user_test.rb
+++ b/test/models/user_test.rb
@@ -26,6 +26,12 @@ class UserTest < ActiveSupport::TestCase
     refute User.new(email: test_email).valid?
   end
 
+  test 'email must be saved downcased' do
+    new_email = 'NEW_EMAIL@exaMple.cOm'
+    assert create(:user, email: new_email.downcase)
+    refute User.new(email: new_email).valid?
+  end
+
   test 'doesnt geocode until we save' do
     u = build(:user, latitude: nil, longitude: nil)
     assert u.valid?
--- a/test/models/user_test.rb
+++ b/test/models/user_test.rb
@@ -26,10 +26,15 @@ class UserTest < ActiveSupport::TestCase
     refute User.new(email: test_email).valid?
   end
 
-  test 'email must be saved downcased' do
-    new_email = 'NEW_EMAIL@exaMple.cOm'
-    assert create(:user, email: new_email.downcase)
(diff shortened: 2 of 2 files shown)


app/models/user.rb
```diff
@@ -103,4 +104,8 @@ def token
   def upcase_state
    state.upcase! if state
   end
+
+  def downcase_email
+    email.downcase! if email
```
hpjaj: 👍

test/models/user_test.rb
```diff
@@ -26,6 +26,12 @@ def setup
     refute User.new(email: test_email).valid?
   end
 
+  test 'email must be saved downcased' do
+    new_email = 'NEW_EMAIL@exaMple.cOm'
+    assert create(:user, email: new_email.downcase)
+    refute User.new(email: new_email).valid?
+  end
+
```
hpjaj: Perhaps add a test that specifically confirms that the email saved to the database is downcased?  Ping me if you'd like any ideas on that.

test/models/user_test.rb
```diff
@@ -26,6 +26,12 @@ def setup
     refute User.new(email: test_email).valid?
   end
 
+  test 'email must be saved downcased' do
+    new_email = 'NEW_EMAIL@exaMple.cOm'
+    assert create(:user, email: new_email.downcase)
+    refute User.new(email: new_email).valid?
```
rickr: I think you can remove this test as it's not really testing the validity of down casing the email. You can test this out by commenting out your newly added `before_save` filter. If email case was affecting this test you would expect it to fail.

sabrown84: @rickr I tried your suggestion to try to get my test to fail and it wouldn't recognize the test at all...it seemed to pass without the `before_save`...so I looked into the config/initializers/devise.rb line 49 downcases emails upon user creation. If I'm understanding this right I'm having trouble testing the `before_save` because devise is already downcasing the emails. Is that right?

test/models/user_test.rb
sabrown84: @hpjaj  I tried commenting out the devise line and changing the test to:
`test 'email must be saved downcased' do
    u = build(:user, email: 'NEW_EMAIL@exaMple.cOm')
    u.save
    assert_equal 'new_email@example.com', u.email
  end`
It seemed to work and pass as a test, but I'm unsure if its needed.

hpjaj: @sabrown84 - You're on the right track.   Here's some pseudo code for how you'd want the test to go:

```
test 'email is downcased on create' do
  address = 'NEW_EMAIL@exaMple.cOm'
  user = create (not build) a user with that address
  reload the user (may or may not need this)
  assert that the user's email == downcased version of address
end
```

That test's the case when a `User` is being created.  A good second test would be when a `User` is being updated.

```
test 'email is downcased on update' do
  user = create a user with that address
  address = 'NEW_EMAIL@exaMple.cOm'
  user's address = address
  save the user
  assert that the user's email == downcased version of address
end
```

[assert_equal](https://apidock.com/ruby/v1_9_3_392/Test/Unit/Assertions/assert_equal) is a good assertion method to use for these.

test/models/user_test.rb
hpjaj: @sabrown84 - You can just remove this test, in favor of the one(s) I described above.

test/models/user_test.rb
sabrown84: Ok thank you so much. I'll make those changes. This makes so much sense, thank you. I really appreciate the feedback.

Issue #76, opened 2017-07-04
Convert current code_schools endpoint generation to a more extendible architecture


kylemh:
# Improvement
## Why is this improvement being suggested?
I noticed that Galvanize is missing some locations, there may be other code schools with new locations or other out-of-date details. There also have been a few code school representatives that want to add their school to our site, but have to deal with manually updating `code_schools.yaml`. Let's implement CRUD functionality by offering an interface for code school reps and OC admins to add/edit/delete code school data.

## What would your improvement do?
Automates/improves the current functionality of adding/updating/deleting code schools from our API.

### NOTE: Information detailed by @hpjaj [here](https://github.com/OperationCode/operationcode_backend/issues/36#issuecomment-312957658)!

- create a new `CodeSchool` db table
- [CRUD](https://en.wikipedia.org/wiki/Create,_read,_update_and_delete) API endpoints for `CodeSchool` records 
- `rake` task that creates a `CodeSchool` record for each entry in `code_schools.yml` 
- wire up the existing `:index` endpoint to fetch `CodeSchool` db records
- test coverage for new endpoints
- update our [API documentation](https://github.com/OperationCode/operationcode_backend/blob/master/apiary.apib)

kylemh:
If this issue is completed before #24 - close that issue.

PrimeTimeTran:
@kylemh I'll try this one if you don't mind me taking it on.

hpjaj:
@PrimeTimeTran - You can ping me if you have any questions on it!

hpjaj:
@PrimeTimeTran - You are welcome to take this one on.  If you choose to do so, be sure to: 

1. set yourself as the **assignee**
2. change the label status from **available** to **in progress**

If for some reason you run into issues making these updates, ping me or @kylemh and we'll update them for you.

PrimeTimeTran:
@hpjaj would you mind giving me a review of what I've got so far? I don't want to make a PR until I have all the tests written as well.
https://github.com/PrimeTimeTran/operationcode_backend

hpjaj:
@PrimeTimeTran - Can you create a pull request in your repository, so that all of your changes are isolated to a branch for code review?  

If you have never created a pull request for an open source project before, this link can give you some guidance on how to do so:  https://mind-on-rails.herokuapp.com/notes/313-free-course-on-how-to-contribute-to-an-open-source-project

hpjaj:
[...]

PR #164, opened 2017-09-22
CodeSchool API Endpoints for CRUD & Rake Task for seeding from code_schools.yml

PrimeTimeTran:
# Description of changes
#76 
Created CRUD API endpoints for the `CodeSchool` model. Also added `Location` model as well because some of the CodeSchools had multiple locations.

Also created rake task for seeding data.
`rake schools:populate`            
Reads from the ./config/code_schools.yml and creates records in the database.


config/database.yml
```diff
@@ -21,7 +21,8 @@ default: &default
   # http://guides.rubyonrails.org/configuring.html#database-pooling
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
   user: postgres
-  host: operationcode-psql
+  # host: operationcode-psql
```
hpjaj: @rickr - thoughts on this change?

lib/tasks/schools_from_yml.rake
```diff
@@ -0,0 +1,30 @@
...
+        name: school["name"], 
+        url: school["url"], 
+        logo: school["logo"], 
+        full_time: school["full_time"],
+        hardware_included: school["hardware_included"],
+        has_online: school["has_online"],
+        online_only: school["online_only"],
+        notes: school["notes"]
+        )
+      object.save
```
hpjaj: Perhaps consider using the `create!` method for creating the `CodeSchool` and `Location` records.  This will prevent any silent failures, and force the raising of an error, should there be an issue.  Then you can `rescue` and `log` those errors.  

Then you'll be able to identify any records that were not created, so you can go back and troubleshoot, manually add them, etc.

Here is [an example](https://github.com/OperationCode/operationcode_backend/blob/master/lib/tasks/resource.rake) of this.

test/controllers/api/v1/code_schools_controller_test.rb
```diff
@@ -0,0 +1,20 @@
...
+    get api_v1_code_schools_path
+    assert_response :success
+  end
+  
+  test ":show will not work for a invalid record" do
+    get api_v1_code_school_path(1)
+    assert_response :missing
+  end
+end
+
```
hpjaj: Perhaps add in some test coverage for the `create`, `update`, and `destroy` actions?  Here are [some examples](https://github.com/OperationCode/operationcode_backend/blob/master/test/controllers/api/v1/team_members_controller_test.rb).

test/factories/code_schools.rb
```diff
@@ -0,0 +1,12 @@
+FactoryGirl.define do
+  factory :code_school do
+    name "CoderSchool"
```
hpjaj: Perhaps leverage the [Faker gem](https://github.com/OperationCode/operationcode_backend/blob/master/test/factories/users.rb) to create unique `name`s, `url`s, `addresses`(in the other `location` factory). 

Here are [some examples](https://github.com/OperationCode/operationcode_backend/blob/master/test/factories/users.rb#L4-L12) of the gem being used.

test/models/code_school_test.rb
```diff
@@ -0,0 +1,15 @@
+require 'test_helper'
+
+class CodeSchoolTest < ActiveSupport::TestCase
+  def setup
```
hpjaj: Since the `code_school.rb` model file does not have any methods in it yet, the only thing you can really test for is validations.  So until we have some methods in there, you can probably just delete this file.

test/models/location_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
+
+class LocationTest < ActiveSupport::TestCase
```
hpjaj: Same thing here.

config/application.rb
```diff
@@ -33,5 +33,9 @@ class Application < Rails::Application
     config.lograge.enabled = true
 
     config.secret_path = Rails.root.join('run/secrets')
+    
+    Raven.configure do |config|
```
hpjaj: Can you elaborate on why this is being added?

app/controllers/api/v1/code_schools_controller.rb
```diff
@@ -2,7 +2,50 @@ module Api
...
+        @school = CodeSchool.find(params[:id])
+        @school.update(params_school)
+        if @school.save
+          render json: @school
+        else 
+          render json: { :errors => @school.errors.full_messages }
+        end
+      end
+      
+      def destroy
```
hpjaj: Please be sure to update our API documentation with these new endpoints (`index, create, show, update, destroy`).  

You'll do so by updating the  [apiary.apib](https://github.com/OperationCode/operationcode_backend/blob/master/apiary.apib) file, following the pattern in the `TeamMember` section.

Here is what that file yields:  http://docs.operationcodeapi.apiary.io/#

config/database.yml
PrimeTimeTran: @hpjaj I just made this change to get it working on my local machine. I'm not sure how I could have gotten it to run the server otherwise.

config/application.rb
PrimeTimeTran: I added this because I got an error `sentry-raven DSN not set`. This was the solution I found on the internet to get the server to run locally. I plan on dropping it when I finish the PR.

app/models/code_school.rb
```diff
@@ -0,0 +1,4 @@
+class CodeSchool < ApplicationRecord
+  validates :name, :url, :logo, :full_time, :hardware_included, :has_online, :online_only, presence: true
```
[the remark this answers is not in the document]
PrimeTimeTran: @hpjaj I'm not sure what OperationCode would like to validate so I figured I just add them all and delete before merging the PR. Concerning this question, yes I did run the rake task locally and the only field that was not on every school was `notes`

PrimeTimeTran:
@hpjaj could you take a look at the API Documentation I've added? Besides this I'll remove the configuration for `Raven` I added in `./config/application.rb` and the change the `host:` back to `host: operationcode-psql` in `./config/database.yml` and I think it'll be close to done. 

Is there anything else you can see that I've missed?

lib/tasks/schools_from_yml.rake
```diff
@@ -0,0 +1,28 @@
+namespace :schools do
+  desc "Reads from the ./config/code_schools.yml and creates records in the database."
+  task populate: :environment do 
+    schools =  YAML::load_file(File.join("./config", "code_schools.yml"), "r")    
+    schools.each do |school|
+      object = CodeSchool.create!(
```
hpjaj: Can you add in a `rescue` block so that we catch any errors, and the rest of the schools that are error free can be created.

test/controllers/api/v1/code_schools_controller_test.rb
```diff
@@ -0,0 +1,51 @@
...
+    school_count = CodeSchool.count + 1
+    get api_v1_code_school_path(school_count), as: :json
+    assert_response :missing
+  end
+  
+  test ":show works for a valid record"do
+    get api_v1_code_school_path(@school), as: :json
+    assert_response :ok
+  end
+
```
hpjaj: Can you add in a test for `:create`.

test/factories/code_schools.rb
hpjaj: Can you update these with `Faker`.

test/factories/locations.rb
```diff
@@ -0,0 +1,11 @@
+FactoryGirl.define do
+  factory :location do
+    va_accepted false
+    address1 "MyString"
```
hpjaj: Here too, removing any `MyString`.

app/controllers/api/v1/code_schools_controller.rb
```diff
@@ -2,7 +2,50 @@ module Api
...
+        render json: CodeSchool.all
+      end
+      
+      def create 
+        school = CodeSchool.new(code_school_params)
+        
+        if school.save 
+          render json: { code_school: school.id }
+        else 
+          render json: { :errors => school.errors.full_messages }
```
hpjaj: Can you change these to the newer `errors: school.errors.full_messages` syntax.

apiary.apib
```diff
@@ -577,3 +577,191 @@ API endpoints that Operation Code's Rails backend makes available to its React f
         {
             errors: "Some error message"
         }
+
+## CodeSchool | Collection [/api/v1/code_schools]
```
hpjaj: Two things:

1.  Please create a free account at apiary.io, and use their editor to confirm that all of these changes do not create any semantic issues 
2. These endpoints are alphabetized, so these additions will need to go at the top of the file

test/factories/locations.rb
PrimeTimeTran: @hpjaj if I use Faker I'd have to add some more code to the tests, do you want me to user faker on all of the attributes?

app/controllers/api/v1/code_schools_controller.rb
PrimeTimeTran: @hpjaj yeah, sorry about that

test/controllers/api/v1/code_schools_controller_test.rb
PrimeTimeTran: @hpjaj coulda swore I wrote one, musta overwritten it somewhere.

test/factories/locations.rb
hpjaj: Adding more code to the tests is fine.  The more representative our tests are of the real world, the better.

Yes, please use faker in situations where uniqueness is valuable, and more representative of the real world.  

You can still choose specific attributes in the tests if needs be, i.e.

```ruby
let(:location) { create :location, address1: "1515 Main St.", .. }
```

PrimeTimeTran: Ok let me do all of this and ping you once I've completed. Thanks for your help~!

PrimeTimeTran:
@hpjaj I've worked on the requests you had.

apiary.apib
```diff
@@ -5,6 +5,195 @@ HOST: https://operationcode.org/
...
+                "has_online": false,
+                "online_only": false,
+                "created_at": "2017-09-21T05:48:50.108Z",
+                "updated_at": "2017-09-21T05:48:50.108Z",
+                "notes": null
+            }
+        ]
+        
+        
+### CodeSchool | Create [/api/v1/code_schools{?name,url,logo,full_time,hardware_included,has_online,online_only,notes}]
```
hpjaj: I believe we'll want this to be an `h2` header instead of an `h3`, with an `h3` beneath it, because it is its own endpoint.

apiary.apib
```diff
@@ -5,6 +5,195 @@ HOST: https://operationcode.org/
...
+                  "code_school": 182
+              }
+
+  + Response 422 (application/json)
+
+          {
+              errors: "Some error message"
+          }
+          
+## CodeSchool | Show [/api/v1/code_schools/{code_school_id}] [GET]
```
hpjaj: We'll want an `h3` header below this one, so that it stands alone as its own endpoint.  

The way apiary enforces formatting, we essentially need to have one `h2` and one `h3` for each new endpoint.

lib/tasks/schools_from_yml.rake
```diff
@@ -0,0 +1,32 @@
...
+            va_accepted: location["va_accepted"],
+            address1: location["address1"],
+            address2: location["address2"],
+            city: location["city"],
+            state: location["state"],
+            zip: location["zip"]
+          )
+        end
+      rescue ActiveRecord::RecordInvalid => e
+        puts "Failed to create CodeSchool: #{e}"
```
hpjaj: Can you add a `Rails.logger.error ...` so that we capture the error in the logs, too.

test/factories/locations.rb
```diff
@@ -0,0 +1,11 @@
+FactoryGirl.define do
+  factory :location do
+    va_accepted Faker::Boolean.boolean
```
hpjaj: In the other factories, the call to `Faker` is wrapped in `{ }`:  https://github.com/OperationCode/operationcode_backend/blob/master/test/factories/users.rb#L4

Can you confirm if these are needed or not?

apiary.apib
PrimeTimeTran: @hpjaj I get complaints from Apiary when I try to format it the way you suggested. 
I don't have another example of a resource with a `show` & `delete` endpoint to reference on how to get by this. 

Do you know of a workaround? Line `92` & `118`

screen shot 2017-09-27 at 12 48 03 am
 so I don't get the semantic issue?

test/factories/locations.rb
PrimeTimeTran: @hpjaj I'm reasonably confident we don't need the `{}`'s, all the tests pass when using the `create(:code_school)` method, where we see the Factory is defined using the same syntax.

lib/tasks/schools_from_yml.rake
PrimeTimeTran: Done.

apiary.apib
hpjaj: For the case where both API endpoints are identical, and have the same parameters, i.e. `[/api/v1/code_schools/{code_school_id}]`, I would: 

- have an `h2` that addresses both (i.e. `## CodeSchool | Show or Delete`)
- have an `h3` for each endpoint, beneath that `h2`

PrimeTimeTran: @hpjaj thanks for the support on completing this task. Mind taking a look again? I've updated with all of your suggestions.

PrimeTimeTran:
@hpjaj I've committed and pushed the changes you requests =).

hpjaj:
@PrimeTimeTran - There ended up being a couple issues with this PR.  I wanted to chat with you about them.

Are you a member of the Operation Code Slack channel?  If yes, what is your @ name?

If no, you can join by going to https://operationcode.org/profile and clicking **Enter our Slack channel**.  

Once you are in Slack, my handle is **@ harry**.

PrimeTimeTran:
@hpjaj I see, sorry I missed them =(.

No, I'm not. ok I'll join =)

PR #165, opened 2017-09-27
Gets data from Meetup API and adds to database in events table

### LINKED ISSUE #40: Pull all meetups from the meetup API
# Feature

1. Get all Operation Code meetup data from Meetup.com's API. 
2. Filter the data and save to the database as events.  Event data will be set up in such a way to expand to other platforms (facebook)
   - Write migration to add source_id, source_updated, and group to existing Events table
   - Compare to existing data to ensure events aren't duplicated. 
   - Compare updated_at (available via Meetup's API) to see if an event has been updated. 
3. Write a rake task to pull the data. 
4. Render data as json at events endpoint. 

leenyburger:
# Description of changes
The purpose of this PR is to:

>Use Meetup's API to retrieve all Operation Code Events and add them to database. 

Creates the following files: 
1. Meetup.rake : A rake task which adds all unique available meetup.com Operation Code events to the database as events. 
2. Meetup.rb 
  - Gets data from the meetup.com API 
  - Error checks the response from Meetup's API to ensure response code = 200
  - Getting all the events is a two step process because of the way the Meetup API returns /pro data. First the groups are returned, then the events can be returned by group. 
  - Extraneous json data is removed from the event before the event is saved in the database.
  - "group" is added as a parameter to the json data. This is not returned in the Meetup json 
  response, it is part of the search parameter. 
  -  Each event has a source_id provided by Meetup.  Unique events are added to the database. Uniqueness is checked against both the source_id and the source ("Meetup") to provide for expansion to other platforms such as Facebook. 
  - If an event is not unique, the source_updated is checked against the existing database entry. If the event was updated on the API side, it is updated in the database.  This is determined using the source_updated field provided by Meetup. 
3. Events_controller : Updated to return json data of Events. 
4. Event.rb : Meetup does not enforce all fields. Several validations were dropped, and the source fields were added. 
5. Events.rb: Updated to properly use factory girl and faker gem
6. Event_test.rb: Updated to remove test for validations that are not longer applicable. 
7. Events_create_test.rb: Integration tests added

# Issue Resolved

Fixes #40


app/models/meetup.rb
```diff
@@ -0,0 +1,95 @@
...
+
+ private 
+	
+  def group_names
+    operationcode_data.map { |group| group["urlname"]  }
+  end
+
+  def event_duration(event)
+    if event['duration'].present? 
+      duration = event['duration'] / 1000
```
hpjaj: Since you are returning the value in both these conditionals, and not using the `duration` assignment, you can remove this assignment in both places.  i.e.

```ruby
if event['duration'].present? 
  event['duration'] / 1000
else
  0
end
```

app/models/meetup.rb
```diff
@@ -0,0 +1,95 @@
...
+    else
+      duration = 0
+    end
+  end
+
+  def return_value_for(response)
+    if response.code.to_i==200
```
hpjaj: Can you add a space around both sides of the `==` operator

app/models/meetup.rb
```diff
@@ -0,0 +1,95 @@
...
+      response.parsed_response
+    else 
+      error_message "Error fetching data from Meetup API"
+      raise error_message
```
hpjaj: Since you are only calling a `raise` here, and not an additional `logger` or `puts`, you can consolidate these two lines into:

```ruby
raise "Error fetching data from Meetup API"
```

db/migrate/20170813163239_add_columns_to_events.rb
```diff
@@ -0,0 +1,8 @@
+class AddColumnsToEvents < ActiveRecord::Migration[5.0]
+  def change
+  	add_column :events, :source_id, :string, index: true, unique: true 
+  	add_column :events, :source, :string, index: true
```
hpjaj: The rails convention for polymorphic associations is for this column to end in `_type`, i.e. `source_type`.  `source_id` and `source_type` go together.

Technically, since we are not applying this is as a polymorphic association (because the `source_id` is a `string` instead of an `integer`), you don't **have to** change it.  But it would be nice :)  If you do, be sure to find everywhere it is used and replace.

lib/tasks/meetup.rake
```diff
@@ -0,0 +1,10 @@
+namespace :meetup do
+	desc "Rake task to get Meetup API data"
+ 	task :fetch => :environment do 
+  	begin
+			Meetup.new.add_events_to_database!
```
hpjaj: You'll need to fix all the indenting in this rake task

test/factories/events.rb
```diff
@@ -1,15 +1,21 @@
 FactoryGirl.define do
   factory :event do
+    trait :updated_in_the_future do
+        source_updated {2.days.from_now}
```
hpjaj: outdent

test/factories/events.rb
```diff
@@ -1,15 +1,21 @@
...
+    end
+
+    trait :updated_today do
+        source_updated { Date.today }
```
hpjaj: here too

test/integration/events_create_test.rb
```diff
@@ -0,0 +1,26 @@
...
+    Event.delete_all
+  end
+
+  test 'valid event information saves' do
+    assert_difference 'Event.count', 1 do
+      create :event, :updated_today
+    end
+  end
+
+  test 'find_or_initialize_by finds duplicate event by source_id and source' do
```
hpjaj: We can remove this test, as it is exclusively testing a Rails method, `find_or_initialize_by`.

test/integration/events_create_test.rb
```diff
@@ -0,0 +1,26 @@
...
+    event_in_database = create :event, :updated_today
+    meetup_event = event_in_database
+    my_event = Event.find_or_initialize_by(source_id: meetup_event[:source_id], source: "Meetup") 
+    assert_equal my_event, meetup_event
+  end
+
+  test 'find_or_initialize_by creates new event if event is not in database' do
```
hpjaj: This one, too.

app/models/meetup.rb
```diff
@@ -0,0 +1,95 @@
+class Meetup
```
hpjaj: We will want to move this file to the `lib` directory, as `Meetup` is not a database backed model, inheriting from `ApplicationRecord `.

db/migrate/20170813163239_add_columns_to_events.rb
mwagz: How big is the events table? Will there be any downtime associated with this PR? @rickr (You might not be the best person to tag for this question)

hpjaj: @mwagz - Good question!  We are good with this merging at any time, no downtime required.  

The `event` table is relatively small.  And when this `rake` task is running daily, @leenyburger has checked and it will be pulling a small number of events in per day.    

@leenyburger , can you ballpark that number for us.

leenyburger: Yup, 20-30 events per day

app/lib/meetup.rb
```diff
@@ -0,0 +1,95 @@
...
+  def add_events_to_database!
+    event_details_by_group.each do |event|
+      my_event = Event.find_or_initialize_by(source_id: event[:source_id], source_type: "Meetup") 
+      if my_event.new_record? || my_event_updated?(my_event)
+        my_event.update!(event)
+      end
+    end
+  end
+
+  def my_event_updated?(my_event)
```
hpjaj: Since this is not being called outside of this class, you can make this a `private` method

app/models/event.rb
```diff
@@ -1,14 +1,6 @@
...
-	validates :start_date, presence: true
-	validates :address1, presence: true
-	validates :city, presence: true
-	validates :state, presence: true, length: {is: 2} #Dropdown selection on front end?  
-	VALID_ZIP_REGEX= /\A\d{5}-?\d{4}?\z/
-	validates :zip, presence: true, format:{with:VALID_ZIP_REGEX}
-	validates :scholarship_available, inclusion: {in: [true, false]}
-	VALID_URL_REGEX = URI::regexp 
+	validates :name, :start_date, :source_updated, :source_type, :address1, :city, presence: true
+	validates :source_id, presence: true, uniqueness: {case_sensitive: false}
```
hpjaj: Can you add spacing around the `{` and `}`

app/lib/meetup.rb
```diff
@@ -0,0 +1,95 @@
+class Meetup
```
hpjaj: We'll want to add test coverage for this class

hpjaj:
@leenyburger - heads up, Travis CI found a failing test

test/lib/meetup_test.rb
```diff
@@ -0,0 +1,15 @@
+require 'test_helper'
+require_relative '../support/meetups/sample_api_responses'
+
+class MeetupTest < ActiveSupport::TestCase
+	test "good data is saved" do
+		Meetup.stubs(:get).with("pro/operationcode/groups", options).returns(group_endpoint_response) #Returns response which includes response.parsed_response and response.code
```
hpjaj: @leenyburger - the issue is that there is a missing `/` in the url.  Should be equal to this url:

https://github.com/leenyburger/operationcode_backend/blob/43341d615958e2d497153bf08a98a7c2d3239d1d/app/lib/meetup.rb#L19

test/lib/meetup_test.rb
```diff
@@ -0,0 +1,90 @@
...
+class MeetupTest < ActiveSupport::TestCase
+
+#The code is structured so /groups is always called before /events. Both responses must be stubbed.
+#Stub data is dependent on this urlname, do not change
+  setup do
+    Event.delete_all
+    Meetup.stubs(:get).with("/pro/operationcode/groups", options).returns(group_endpoint_response)
+    @urlname = "Operation-Code-Hampton-Roads"
+    Meetup.stubs(:get).with("/#{@urlname}/events", options).returns(event_endpoint_response)
+    @event1 = {:source_id=>"243869029",
```
leenyburger: I added this because I was trying to update the events for testing with the event created by Faker which didn't work because Faker is class event and to update a db record I need a hash.

test/lib/meetup_test.rb
```diff
@@ -0,0 +1,90 @@
...
+  	assert_raises( Exception ) { Meetup.new.get_events_for(urlname) }
+  end
+
+  #There are 4 events in the stub data, 1 with no venue
+  test "Meetup events with no venue are not included" do
+  	events = Meetup.new.event_details_by_group
+  	assert_equal( events.length, 3 )
+  end
+
+  #Event endpoint stubbed reponse has 3 good event records.
```
leenyburger: add_events_to_database! calls event_details_by_group, which is returning nil in this test.  Same issue as above

test/lib/meetup_test.rb
```diff
@@ -0,0 +1,90 @@
...
+  test "good data saves as new database record" do
+    assert_difference 'Event.count', 3 do
+      Meetup.new.add_events_to_database!
+    end
+  end
+
+  test "duplicate data does not create new record" do
+    dup_event = @event1
```
leenyburger: I got a bit confused between Event objects and hashes to update event objects. Here I use @event1 which is a hash defined above.  Not sure if I should have used a params variable with permitted params.  I didn't because there is nothing coming from a user.  All tests are failing because of the add_events_to_database! error mentioned above.

test/lib/meetup_test.rb
```diff
@@ -0,0 +1,90 @@
...
+  test "event response code of 200 returns response" do
+  	assert_nothing_raised { Meetup.new.get_events_for(@urlname) }
+  end
+
+  test "event response code not equal to 200 raises error" do
+  	Meetup.stubs(:get).with("/#{@urlname}/events", options).returns(build_response(400))
+  	assert_raises( Exception ) { Meetup.new.get_events_for(urlname) }
+  end
+
+  #There are 4 events in the stub data, 1 with no venue
```
leenyburger: All the tests from here on are throwing an error because there is an issue with the group_names which is called from event_details_by_group.  The groups are returning as nil.  
It runs fine in the rake task.  
As far as I can see, the difference is in def group_names. When running the rake task, operationcode_data returns a hash "name"=>"Come Learn To Code!".  When this data is mapped using "urlname", the groups render fine.  When the data is mapped while running the test, the groups map as nil. When I use byebug to inspect this during the test, operationcode_data is a hash of the form :name => "Come Learn To Code!".  I copied the stub data directly from the console, so I don't know where the alternate hash syntax came from. Also, does it matter?

test/lib/meetup_test.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  test "duplicate data does not create new record" do
+    dup_event = @event1
+    events = [ @event1, dup_event ] #This should be an array of hashes
+    Meetup.stubs(:event_details_by_group).returns(events)
+    Meetup.new.add_events_to_database!
+    assert_equal 1, Event.count
+  end
+
+  test "updated data is updated in database" do
```
leenyburger: I know there are probably too many asserts in this test, I was just trying to step through it logically.  Will clean up once the group issue is fixed.

test/lib/meetup_test.rb
```diff
@@ -0,0 +1,90 @@
...
+  test "good data saves as new database record" do
+    assert_difference 'Event.count', 3 do
+      Meetup.new.add_events_to_database!
+    end
+  end
+
+  test "duplicate data does not create new record" do
+    dup_event = @event1
+    events = [ @event1, dup_event ] #This should be an array of hashes
+    Meetup.stubs(:event_details_by_group).returns(events)
```
leenyburger: Not sure if I am stubbing event_details_by_group correctly.  It's an instance method, not a class method.  Should it be Meetup.any_instance.stubs?

test/support/meetups/sample_api_responses.rb
```diff
@@ -0,0 +1,317 @@
...
+    "duration"=>7200000,
+    "id"=>"243655210",
+    "name"=>"Come learn to code!",
+    "status"=>"upcoming",
+    "time"=>1516318200000,
+    "updated"=>1506361383000,
+    "utc_offset"=>-18000000,
+    "waitlist_count"=>0,
+    "yes_rsvp_count"=>2,
+    # "venue"=>
```
leenyburger: I commented this out so I can test that events without venue information do not save.

app/lib/meetup.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def my_event_updated?(my_event, event)
+    my_event[:source_updated] < event[:source_updated]
+  end
+
+ private
+
+  def group_names
+    operationcode_data.map { |group| group["urlname"] }
```
leenyburger: Tests are blowing up here because urlname isn't mapping properly.  Works fine when running the rake task.

app/lib/meetup.rb
```diff
@@ -0,0 +1,95 @@
...
+  #Return all meetups based on group names if venue is present. Return array of all op-code events
+  def event_details_by_group
+    events = []
+    group_names.each do |group|
+      get_events_for(group).each do |event|
+        next unless event['venue'].present?
+        events << details_for(event)
+      end
+    end
+    events.flatten
```
leenyburger: Trying to stub event_details_by_group to better test add_events_to_database!  Want to test update cases and duplicates.  event_details_by_group returns an array of hashes.

test/lib/meetup_test.rb
```diff
@@ -0,0 +1,118 @@
...
+    assert_equal 3, Event.count
+  end
+
+  test "updated data is updated in database" do
+    url = "Operation-Code-Hampton-Roads"
+
+    Meetup.stubs(:get).with("/pro/operationcode/groups", options).returns(group_endpoint_response)
+    Meetup.stubs(:get).with("/#{url}/events", options).returns(event_endpoint_response)
+
+    # create a Event record from first item in `event_endpoint_parsed_response`
```
hpjaj: For lines 89 - 102 or so, you can create a method to abstract this logic into.

app/lib/meetup.rb
```diff
@@ -0,0 +1,96 @@
...
+  def add_events_to_database!
+    event_details_by_group.each do |event|
+      saved_event = Event.find_or_initialize_by(source_id: event[:source_id], source_type: "Meetup")
+      if saved_event.new_record? || saved_event_updated?(saved_event, event)
+        saved_event.update!(event)
+      end
+    end
+  end
+
+  def saved_event_updated?(saved_event, event)
```
hpjaj: If this method is not being called outside of this class, we can move it down in the file and make it a `private` method.

app/models/event.rb
```diff
@@ -1,14 +1,6 @@
...
-	validates :description, presence: true
-	validates :start_date, presence: true
-	validates :address1, presence: true
-	validates :city, presence: true
-	validates :state, presence: true, length: {is: 2} #Dropdown selection on front end?  
-	VALID_ZIP_REGEX= /\A\d{5}-?\d{4}?\z/
-	validates :zip, presence: true, format:{with:VALID_ZIP_REGEX}
-	validates :scholarship_available, inclusion: {in: [true, false]}
-	VALID_URL_REGEX = URI::regexp 
+	validates :name, :start_date, :source_updated, :source_type, :address1, :city, presence: true
```
hpjaj: Since new columns are being added, the existing ones will not have data in those newly validated columns.  Which could cause an issue.

Typically the migration would also deal with resolving these issues.  But perhaps in this case we should just remove any validation of `presence` for any newly created columns.

app/models/event.rb
hpjaj: Instead of validating for uniqueness, we would want to verify for a unique combination between the _id and _type.  Like this:  https://github.com/OperationCode/operationcode_backend/blob/master/app/models/git_hub_statistic.rb#L9

Reason being, we could in theory have the exact same id from another source (like facebook); however, we could not have two of the same id from the *same* source.

app/lib/meetup.rb
```diff
@@ -0,0 +1,97 @@
...
+      saved_event = Event.find_or_initialize_by(source_id: event[:source_id], source_type: "Meetup")
+      if saved_event.new_record? || saved_event_updated?(saved_event, event)
+        saved_event.update!(event)
+      end
+    end
+  end
+
+ private
+
+
```
hpjaj: small nit.  pls remove extra space

test/lib/meetup_test.rb
hpjaj: this still needs to be addressed

app/models/event.rb
leenyburger: There shouldn't be any existing data.  I wrote the event.rb endpoint, and when it was originally created the plan was to allow operation code community leaders to add data on the front end.  This was never implemented, and that's when we decided to get data from Meetup and not allow events to be manually entered.  So we should be starting fresh with event data once this is merged.

test/lib/meetup_test.rb
leenyburger: I took the logic that was there and made the create_and_save_old_event method which I put at the end of the file.  Is that not what you meant?

app/models/event.rb
hpjaj: Gotcha.  Can you please confirm with @hollomancer that `Event.count == 0` in prod.  If yes, we're set on the validations.

test/lib/meetup_test.rb
hpjaj: Nice 👍

apiary.apib
```diff
@@ -189,6 +189,35 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
+                "scholarship_available": null,
+                "created_at": "2017-10-16T15:18:12.711Z",
+                "updated_at": "2017-10-16T15:18:12.711Z",
+                "source_id": "243869029",
+                "source_type": "Meetup",
+                "source_updated": "49724-07-09T02:30:00.000Z",
+                "group": "MilSpouseCoders - Robins AFB"
+            }
+        ]
+
```
hpjaj: As you've accounted for the "if things go wrong case" here:  https://github.com/OperationCode/operationcode_backend/pull/165/files#diff-1d8ee09392f941bd379c0eaeeafc6929R7

We'll want to add in what the end user would expect to receive, if things go wrong.  For example:  https://github.com/OperationCode/operationcode_backend/blob/master/apiary.apib#L235-L239

app/models/event.rb
```diff
@@ -1,6 +1,7 @@
 class Event < ApplicationRecord
-	validates :name, :start_date, :source_updated, :source_type, :address1, :city, presence: true
-	validates :source_id, presence: true, uniqueness: {case_sensitive: false}
+	validates :name, :start_date, :source_updated, :source_id, :source_type, :address1, :city, presence: true
+	validates :source_id, uniqueness: { scope: :source_type }
```
hpjaj: 👍

app/models/event.rb
hollomancer: Confirmed!

Issue #191, opened 2017-10-30
Allow database host to be set by environment variable


ohaiwalt:
# Feature
## Why is this feature being added?

In order to set up a staging environment or any other environment not already accounted for, the database host needs to be configurable at runtime.

## What should your feature do?

Allow an environment variable `POSTGRES_HOST` to be passed in to determine the location of the database server. This could potentially default to the current values, or additionally update docker-compose so dev doesn't break.

Depending on implementation, this may be blocked by https://github.com/OperationCode/operationcode_infra/issues/23

PR #194, opened 2017-11-05
dynamic-sidekiq-db-name

hpjaj:
# Description of changes

Sets the `sidekiq` database URL to be dynamic with a named fallback


ohaiwalt:
Supporting documentation [here](https://github.com/mperham/sidekiq/wiki/Advanced-Options#rails-32-and-newer).

LGTM. Related to #191

PR #199, opened 2017-11-20
Social Logins Backend

PixelWitch:
# Description of changes
Preliminary implementation of Social logins, currently undergoing bugfixing and testing


sethbergman:
Crossposting this here:
https://github.com/OperationCode/operationcode_frontend/pull/743#issuecomment-345942238

app/controllers/api/v1/users_controller.rb
```diff
@@ -11,7 +11,6 @@ def index
 
       def create
         user = User.new(user_params)
-
```
hpjaj: Small nit.  Please keep this space in between the group of assigning variables, and taking actions.

app/models/user.rb
```diff
@@ -98,6 +98,21 @@ def token
     JsonWebToken.encode(user_id: self.id, roles: [], email: self.email, verified: verified)
   end
 
+  def self.from_social(data)
+    Rails.logger.info "************ email is = #{data[:email]}"
+    user = User.where(email: data[:email]).first
+
+    # Uncomment the section below if you want users to be created if they don't exist
+    unless user
+         user = User.new(first_name: data[:first_name],
```
hpjaj: Please outdent here to have consistent 2-space indentation

app/controllers/api/v1/users_controller.rb
```diff
@@ -29,6 +29,51 @@ def update
         end
       end
 
+      def exist
+        userData = params[:user]
+        Rails.logger.info "************ exist is = #{userData[:email]}"
+        user = User.where(email: userData[:email]).first
```
hpjaj: Since the `userData` is only being used once, you can save the extra variable assignment, and go straight to:

```ruby
user = User.where(email: params[:user][:email]).first
```

And a cool alternative to the `.where(...).first` pattern is the new'ish `.find_by`.  It either returns the singular record in question, or `nil`, if none is found.

```ruby
user = User.find_by(email: params[:user][:email])
```

This all assumes that we will **always** have a `:user` and `:email` populated in the params.

app/controllers/api/v1/users_controller.rb
```diff
@@ -29,6 +29,51 @@ def update
...
+        Rails.logger.info "************ user is = #{user}"
+        @redirect_path = '/profile' #change this to actually redirect to the social login function
```
hpjaj: Do we have a route established for this, so we can use the dynamic `_path` methods here (i.e. `user_profile_path`, etc.), instead of hard coding with a string?

app/controllers/api/v1/users_controller.rb
```diff
@@ -29,6 +29,51 @@ def update
...
+        unless user
+             @redirect_path = '/additional_info'
```
hpjaj: Here too.

Also, can you outdent 2 spaces.

app/controllers/api/v1/users_controller.rb
```diff
@@ -29,6 +29,51 @@ def update
...
+        end
+        Rails.logger.info "!!!!!!!!!!!! path is #{@redirect_path}"
+        render json: {
```
hpjaj: Small nit.  Since we are only returning one `key: value` pair, perhaps reduce this down to one line.  i.e.

```ruby
render json: { redirect_to: @redirect_path }
```

app/controllers/api/v1/users_controller.rb
```diff
@@ -29,6 +29,51 @@ def update
         end
       end
 
+      def exist
```
hpjaj: Can you please add in some [yard documentation](http://www.rubydoc.info/gems/yard/file/docs/Tags.md) on these new methods.  

Specifically outlining:

- what the method does
- what params should look like; what is it expecting
- a `@see` to any relevant link that would help future developers understand the method, if applicable

Here are some examples of that:

- https://github.com/OperationCode/operationcode_backend/blob/master/app/models/user.rb#L62-L70
- https://github.com/OperationCode/operationcode_backend/blob/master/app/models/resource.rb#L8-L11

-
 https://github.com/OperationCode/operationcode_backend/blob/master/app/models/git_hub_statistic.rb#L21-L25

app/controllers/api/v1/users_controller.rb
```diff
@@ -29,6 +29,51 @@ def update
...
+        Rails.logger.info "!!!!!!!!!!!! path is #{@redirect_path}"
+        render json: {
+          redirect_to: @redirect_path
+        }
+      end
+
+      def social
+        Rails.logger.info "************ got to create = #{params.inspect}"
+          path = ''
+          arr = User.from_social(params[:user])
```
hpjaj: Can we abstract lines 50 - 52 out into a private method?  For example, `set_social_user`, or something like that?

app/models/user.rb
```diff
@@ -98,6 +98,25 @@ def token
     JsonWebToken.encode(user_id: self.id, roles: [], email: self.email, verified: verified)
   end
 
+  def self.from_social(data)
+    Rails.logger.info "************ email is = #{data[:email]}"
+    user = User.where(email: data[:email]).first
+
+    path = '/profile'
```
hpjaj: Same here, re: `_path` methods

app/models/user.rb
```diff
@@ -98,6 +98,25 @@ def token
...
+    unless user
+      Rails.logger.info "!!!!!! path is = #{path}"
+      user = User.new(first_name: data[:first_name],
```
hpjaj: Can we carry this first attr down one line, for consistency and readability, like this:

```ruby
user = User.new(
  first_name: data[:first_name],
  last_name: data[:last_name],
  email: data[:email],
  zip: data[:zip],
  password: data[:password]
)
```

app/models/user.rb
```diff
@@ -98,6 +98,25 @@ def token
...
+        last_name: data[:last_name],
+        email: data[:email],
+        zip: data[:zip],
+        password: data[:password]
+      )
+      path = '/signup-info'
+    end
+
+    arr = [user, path]
```
hpjaj: You can remove this `arr` variable assignment.  Ruby will actually return the last line, for you.

So you can change it to just:

```ruby
[user, path]
```

PixelWitch: done.

app/models/user.rb
PixelWitch: done.

app/controllers/api/v1/users_controller.rb
PixelWitch: done.

app/controllers/api/v1/users_controller.rb
PixelWitch: I wasn't able to find a _path for this, but I've been thinking about it, and I don't see how a Rails path would work anyway, since the actual location change is taking place in React on the frontend. @redirect path is just getting sent back to the frontend, it's not used to go anywhere in the backend.

PixelWitch: does react have dynamic path methods?

app/controllers/api/v1/users_controller.rb
PixelWitch: outdenting is done.

PixelWitch: For the redirect_path see above.

app/controllers/api/v1/users_controller.rb
PixelWitch: done.

app/controllers/api/v1/users_controller.rb
PixelWitch: done.

app/controllers/api/v1/users_controller.rb
PixelWitch: done, let me know if you need more / different info though.

app/controllers/api/v1/users_controller.rb
hpjaj: np, the strings will work

app/controllers/api/v1/users_controller.rb
```diff
@@ -29,6 +29,57 @@ def update
...
       end
 
+      # exist
+      #
+      # When given a user's email, checks if that user is already registered in the database
+      # and sends the proper redirect url to the frontend accordingly. Called when logging in
+      # via social media.
+      # @param params The data passed in from the frontend, from which [:user][:email] (the user's email) is extracted. Expecting :user in params with parameter :email inside.
+
+      def exist
```
hpjaj: Yeah, Yard documentation takes a while to get the hang of.  For this particular method, perhaps something like:

```ruby
      # For social media logins, renders the appropriate `redirect_to` path.
      #
      # @param params [Hash] ActionController::Parameters that require a :user key, with a nested :email key.
      #   For example: { user: { email: "john@example.com" } }
      # @return [String] A string of the user's redirect_to path
      #
      def exist
        ...
      end
```

app/controllers/api/v1/users_controller.rb
```diff
@@ -29,6 +29,57 @@ def update
...
+        Rails.logger.info "!!!!!!!!!!!! path is #{@redirect_path}"
+        render json: { redirect_to: @redirect_path }
+      end
+
+      # social
+      #
+      # When using social logins, creates the user in the database if necessary, then logs them in
+      # and sends the proper redirect url to the frontend accordingly.
+
+      def social
```
hpjaj: You wanna give another stab at this Yard doc, based on the above?

app/controllers/api/v1/users_controller.rb
```diff
@@ -50,6 +101,17 @@ def by_location
 
       private
 
+      # set_social_user
```
hpjaj: I failed to mention this, but typically only `public` methods are documented.  Unless they are confusing, or justify documentation for future developer understanding.  

So you can remove the docs on this one.

app/controllers/api/v1/users_controller.rb
```diff
@@ -50,6 +101,17 @@ def by_location
...
+      #
+      # When given a user's calls the method to create the user if needed, which returns the user and the redirect path.
+      # @param params The data passed in from the frontend, from which the user's data is extracted. Expecting :user in params.
+
+      def set_social_user
```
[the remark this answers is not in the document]
PixelWitch: done.

app/controllers/api/v1/users_controller.rb
PixelWitch: done.

app/controllers/api/v1/users_controller.rb
PixelWitch: done.

app/controllers/api/v1/users_controller.rb
PixelWitch: Yep, I've revised the docs!

Makefile
```diff
@@ -27,6 +27,10 @@ build:
 console:
 	docker-compose run ${RAILS_CONTAINER} rails console
 
+.PHONY: db_create
```
hpjaj: Please change this to `routes` instead of `db_create`

PixelWitch: done.

test/controllers/api/v1/users_controller_test.rb
```diff
@@ -21,6 +21,46 @@ class Api::V1::UsersControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
   end
 
+  test "exist returns /profile if the user is already registered" do
+    bob = create(:user, first_name: 'Bob', last_name: 'Belcher', email: 'bobs@burgers.com', zip: '62149', password: 'Archer' )
+    params = { user: { email: 'bobs@burgers.com'} }
+    post api_v1_users_exist_url, params: params, as: :json
```
hpjaj: With the tests, a testing pattern we try to follow is to separate and group the _setup_, _actions_, and _expectations_.  

The way that would look for this test would be:

```ruby
test "exist returns /profile if the user is already registered" do
  bob = create(:user, first_name: 'Bob', last_name: 'Belcher', email: 'bobs@burgers.com', zip: '62149', password: 'Archer' )
  params = { user: { email: 'bobs@burgers.com'} }

  post api_v1_users_exist_url, params: params, as: :json
  json = response.parsed_body
  
  assert_equal '/profile', json['redirect_to']
end
```

Can you apply that testing pattern to all your specs.

PixelWitch: done.

hpjaj:
@PixelWitch - Looks great 👍   Once the frontend PR has signed off, and the `Rails.logger`s, etc., has been removed, this will be approved.

app/controllers/api/v1/users_controller.rb
```diff
@@ -29,6 +29,49 @@ def update
...
+      # then logs them in, and renders the appropriate `redirect_to` path depending on whether the user is logging in
+      # for the first time.
+      #
+      # @return [String] A string of the user's redirect_to path
+      # @return [Json] A serialied JSON object derived from current_user
+      # @return [Token] A token that the frontend stores to know the user is logged in
+      # @see https://github.com/OperationCode/operationcode_backend/blob/master/app/controllers/api/v1/sessions_controller.rb#L8-L20
+      #
+      def social
+          path = ''
```
hpjaj: this can be removed

PixelWitch: done.

hpjaj:
@PixelWitch - You'll need to clean up the commits.  This is reflecting 129 commits, from multiple contributors.

kylemh:
Shucks. I suggested to do a rebase. I'll show her how to squash before it's due time.

kylemh:
@hpjaj you mentioned squashing commits at 100+, but I only see 41 - would you like them to be cleaned up a bit still for a good commit log?

hpjaj:
@kylemh - @PixelWitch was able to resolve commingled commits here https://github.com/OperationCode/operationcode_backend/pull/199/commits/9b1496120574ac3a3e63add7e749a15593bcf837 .  Nah I'm fine with the commit history as is.

app/controllers/api/v1/users_controller.rb
```diff
@@ -29,6 +29,48 @@ def update
...
+      # Requires that the ActionController::Parameters contain a :user key, with a nested :email key.
+      # For example: { user: { email: "john@example.com" } }
+      #
+      # @return [String] A string of the user's redirect_to path
+      # @see https://github.com/zquestz/omniauth-google-oauth2#devise
+      #
+      def exist
+        user = User.find_by(email: params[:user][:email])
+
+        @redirect_path = '/profile'
```
hpjaj: Since we are not passing `@redirect_path` to a view, can we switch it to a local variable (i.e. `redirect_path`).

app/controllers/api/v1/users_controller.rb
```diff
@@ -29,6 +29,48 @@ def update
...
+        unless user
```
hpjaj: Since `find_by` returns `nil` if the record is not found, can we flip this conditional to read in the affirmative, for readability, i.e.:

```ruby
if user.nil?
  ...
```

app/controllers/api/v1/users_controller.rb
```diff
@@ -29,6 +29,48 @@ def update
...
+      # then logs them in, and renders the appropriate `redirect_to` path depending on whether the user is logging in
+      # for the first time.
+      #
+      # @return [String] A string of the user's redirect_to path
+      # @return [Json] A serialied JSON object derived from current_user
+      # @return [Token] A token that the frontend stores to know the user is logged in
+      # @see https://github.com/OperationCode/operationcode_backend/blob/master/app/controllers/api/v1/sessions_controller.rb#L8-L20
+      #
+      def social
+          set_social_user_and_redirect
```
hpjaj: Instead of setting instance variables here for `@user` and `@redirect_path` through the `set_social_user_and_redirect` method, which could potentially leak elsewhere in the class, you can have this `User.from_social(params[:user])` method set the two local variable directly here:

```ruby
user, redirect_path = User.from_social(params[:user])
```

app/controllers/api/v1/users_controller.rb
```diff
@@ -50,6 +92,12 @@ def by_location
 
       private
 
+      def set_social_user_and_redirect
+        user_and_redirect = User.from_social(params[:user])
```
hpjaj: Based on the comment about, you can delete this method.

app/models/user.rb
```diff
@@ -104,6 +104,38 @@ def token
...
+  # For example: { user: {first_name: "Jane"},{last_name: "Doe"},{ email: "jane@example.com" },{ zip: "12345"},{ password: "Th!sIs@Pa22w0rd"} }
+  #
+  # @param data The data passed in from the frontend, from which the user's info is extracted.
+  # @return [String] A string of the user's redirect_to path
+  # @return [Json] A serialied JSON object derived from current_user
+  # @return [Token] A token that the frontend stores to know the user is logged in
+  # @return [user, path] The user and the redirect path, in an array
+  # @see https://github.com/zquestz/omniauth-google-oauth2#devise
+  #
+  def self.from_social(data)
```
hpjaj: As this returns both a `User` record, and a `path`, perhaps rename this method to be in keeping with that.  Perhaps `fetch_social_user_and_redirect_path`.

app/models/user.rb
```diff
@@ -104,6 +104,38 @@ def token
...
+    user = User.find_by(email: data[:email])
+
+    path = '/profile'
+    unless user
```
hpjaj: You can flip this conditional to the affirmative, as well.

app/controllers/api/v1/users_controller.rb
```diff
@@ -29,6 +29,48 @@ def update
...
+      #
+      # Requires that the ActionController::Parameters contain a :user key, with a nested :email key.
+      # For example: { user: { email: "john@example.com" } }
+      #
+      # @return [String] A string of the user's redirect_to path
+      # @see https://github.com/zquestz/omniauth-google-oauth2#devise
+      #
+      def exist
+        user = User.find_by(email: params[:user][:email])
+
```
hpjaj: Nit, can you remove this empty space so that the variable assignments are grouped together, and separated from the conditional with a space.

app/controllers/api/v1/users_controller.rb
```diff
@@ -29,6 +29,48 @@ def update
...
+      # For social media logins, renders the appropriate `redirect_to` path depending on whether the user is registered.
+      #
+      # Requires that the ActionController::Parameters contain a :user key, with a nested :email key.
+      # For example: { user: { email: "john@example.com" } }
+      #
+      # @return [String] A string of the user's redirect_to path
+      # @see https://github.com/zquestz/omniauth-google-oauth2#devise
+      #
+      def exist
+        user = User.find_by(email: params[:user][:email])
```
hpjaj: Cool method on hash is [dig](http://ruby-doc.org/core-2.3.0_preview1/Hash.html#method-i-dig).

This will allow you to fetch a key that is nested in a hash, while simultaneously protecting against any upstream keys being `nil`.

For example, if for some reason `params[:user]` was `nil`, this would blow up.

But with `dig`, you can call:

```ruby
params.dig(:user, :email)
```

And if `:user` is not present, the call would return `nil`, and not blow up.  It is kind of like `:try`, for hashes.

app/controllers/api/v1/users_controller.rb
```diff
@@ -29,6 +29,48 @@ def update
...
+            render json: {
+              token: @user.token,
+              user: UserSerializer.new(current_user),
+              redirect_to: @redirect_path
+            }
+          else
+            redirect_to new_user_registration_url, alert: @user.errors.full_messages.join("\n")
+          end
+      end
+
```
hpjaj: A best practice I often hear DHH, and many others, recommending that **all** controller actions be RESTful.  Meaning, trying to use only CRUD actions (`index/create/new/edit/show/update/destroy`), unless you cannot do so.  

This would require an architectural change in that these two new controller actions would go into a new, dedicated controller.  Regarding the naming of the controller, you'd try to come up with one where the standard RESTful actions  make sense.

Perhaps something like `social_users_controller.rb`.

In this way, the `exist` action could be the `show` action.  And the `social` action could be the `create` action.

And then you could change the one-off routes to be resourceful routes (i.e. `resources :social_users, only: [:create, :show]`

config/routes.rb
```diff
@@ -16,6 +16,8 @@
 
       get '/users/by_location', to: 'users#by_location'
       post '/users/profile/verify', to: 'users#verify'
+      post '/users/social', to: 'users#social'
```
hpjaj: See comments in `users_controller`

test/controllers/api/v1/users_controller_test.rb
```diff
@@ -21,6 +21,54 @@ class Api::V1::UsersControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
   end
 
+  test "exist returns /profile if the user is already registered" do
```
hpjaj: These tests would be moved to a new controller test file.

app/controllers/api/v1/users_controller.rb
PixelWitch: done.

app/controllers/api/v1/users_controller.rb
PixelWitch: done.

app/controllers/api/v1/users_controller.rb
PixelWitch: done.

app/controllers/api/v1/users_controller.rb
PixelWitch: done.

app/models/user.rb
PixelWitch: done.

app/models/user.rb
PixelWitch: done.

app/controllers/api/v1/users_controller.rb
PixelWitch: done, also grouped my other methods the same way.

app/controllers/api/v1/users_controller.rb
PixelWitch: done, and changed to this pattern for the other key fetches.

hpjaj:
@PixelWitch - How goes with the above code review suggestions?

app/controllers/api/v1/social_users_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+      # For social media logins, renders the appropriate `redirect_to` path depending on whether the user is registered.
+      #
+      # Requires that the ActionController::Parameters contain a :user key, with a nested :email key.
+      # For example: { user: { email: "john@example.com" } }
+      #
+      # @return [String] A string of the user's redirect_to path
+      # @see https://github.com/zquestz/omniauth-google-oauth2#devise
+      #
+      def show
+        user = User.find_by(email: params.dig(:id))
```
hpjaj: As the `email` key is nested in the `user` key, this would be `(email: params.dig(:user, :email))`

app/models/user.rb
```diff
@@ -104,6 +104,38 @@ def token
...
+  #
+  # @param data The data passed in from the frontend, from which the user's info is extracted.
+  # @return [String] A string of the user's redirect_to path
+  # @return [Json] A serialied JSON object derived from current_user
+  # @return [Token] A token that the frontend stores to know the user is logged in
+  # @return [user, path] The user and the redirect path, in an array
+  # @see https://github.com/zquestz/omniauth-google-oauth2#devise
+  #
+  def self.fetch_social_user_and_redirect_path(data)
+    user = User.find_by(email: data.dig(:email))
```
hpjaj: Same here, assuming email is nested in user

app/models/user.rb
```diff
@@ -104,6 +104,38 @@ def token
...
+
+    path = '/profile'
+    if user.nil?
+      user = User.new(
+        first_name: data.dig(:first_name),
```
hpjaj: And all these.  Take a look through here so you have a better feeling for the `dig` method:  http://ruby-doc.org/core-2.3.0/Hash.html#method-i-dig

app/controllers/api/v1/users_controller.rb
```diff
@@ -29,6 +29,48 @@ def update
...
 
+      # For social media logins, renders the appropriate `redirect_to` path depending on whether the user is registered.
+      #
+      # Requires that the ActionController::Parameters contain a :user key, with a nested :email key.
+      # For example: { user: { email: "john@example.com" } }
+      #
+      # @return [String] A string of the user's redirect_to path
+      # @see https://github.com/zquestz/omniauth-google-oauth2#devise
+      #
+      def exist
```
hpjaj: Reminder to delete all the code from this file that your ported over to the new controller

config/routes.rb
```diff
@@ -16,12 +16,17 @@
...
+      post '/users/exist', to: 'users#exist'
```
hpjaj: Reminder to delete all the routes outside of the new `social_users` ones

app/controllers/api/v1/social_users_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+      # For social media logins, renders the appropriate `redirect_to` path depending on whether the user is registered.
+      #
+      # Requires that the ActionController::Parameters contain a :user key, with a nested :email key.
+      # For example: { user: { email: "john@example.com" } }
+      #
+      # @return [String] A string of the user's redirect_to path
+      # @see https://github.com/zquestz/omniauth-google-oauth2#devise
+      #
+      def show
+        logger.info(email: params.dig(:email));
```
hpjaj: can you pls remove these loggers

test/factories/code_schools.rb
```diff
@@ -8,6 +8,6 @@
     has_online Faker::Boolean.boolean
     online_only Faker::Boolean.boolean
     notes Faker::Lorem.paragraph
-    mooc Faker::Boolean.boolean
+    mooc true
```
hpjaj: I just merged a PR that fixes this issue:  https://github.com/OperationCode/operationcode_backend/pull/261

Please rebase on master, and then change this to `mooc false`, and retest.  You should be good to go at that point.

app/controllers/api/v1/social_users_controller.rb
PixelWitch: done!

test/factories/code_schools.rb
PixelWitch: and done! tests check out, I changed mooc back to a faker boolean.

app/controllers/api/v1/social_users_controller.rb
```diff
@@ -0,0 +1,55 @@
+module Api
+  module V1
+    class SocialUsersController < ApplicationController
+      before_action :authenticate_user!, only: %i[update]
+
+      # For social media logins, renders the appropriate `redirect_to` path depending on whether the user is registered.
+      #
+      # Requires that the ActionController::Parameters contain a :user key, with a nested :email key.
+      # For example: { user: { email: "john@example.com" } }
```
hpjaj: Is this still what the params look like?  I ask b/c `params.dig(:email)` implies that `email` is no longer nested in a `user` hash.

app/models/user.rb
```diff
@@ -105,18 +105,49 @@ def token
     JsonWebToken.encode(user_id: self.id, roles: [], email: self.email, verified: verified)
   end
 
+  # For social media logins, When given a user's info creates their database entry if they do not already have one, and
+  # sets the redirect path for the frontend to go to after the user is logged in
+  # for the first time.
+  #
+  # Requires that all data contains :first_name, :last_name, :email, :zip, and :password keys.
+  # For example: { user: {first_name: "Jane"},{last_name: "Doe"},{ email: "jane@example.com" },{ zip: "12345"},{ password: "Th!sIs@Pa22w0rd"} }
```
hpjaj: This is an invalid structure, with all of these hashes separated by commas.  Can you pls update.

app/controllers/api/v1/social_users_controller.rb
PixelWitch: done, changed it to reflect the new input format.

app/models/user.rb
PixelWitch: I took out the commas, let me know if what it is now is what you're looking for in terms of format.

hpjaj:
@PixelWitch - Looking good!  Can you pls add in the API docs for the new endpoints.  Steps for that are described in here:  https://github.com/OperationCode/operationcode_backend#api-documentation

apiary.apib
```diff
@@ -1043,6 +1043,97 @@ API endpoints that Operation Code's Rails backend makes available to its React f
             errors: "Some error message"
         }
 
+## Social User | Create [/api/v1/social_users{email,first_name,last_name,zip,password}]
```
hpjaj: There needs to be a `?` after the `{` for apiary to recognize all the param individually.  Here is an example:  https://github.com/OperationCode/operationcode_backend/blob/master/apiary.apib#L43

PixelWitch: done!

apiary.apib
```diff
@@ -1043,6 +1043,97 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
+                "/login"
+            }
+
++ Response 500 (application/json)
+
+        {
+            errors: "Some error message"
+        }
+
+## Social User | Show [/api/v1/social_users{email}]
```
hpjaj: Same formatting requirements apply here, too

PixelWitch: done.

app/controllers/api/v1/social_users_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+          else
+            redirect_to new_user_registration_url, alert: @user.errors.full_messages.join("\n")
+          end
+      end
+
+      private
+
+      def social_user_params
+        params.require(:user).permit(
+          :email
```
hpjaj: Here is the request body for the `create` endpoint (from the apiary docs):

```
{
    "user":
        {
            "email": "jon@doe.com",
            "first_name": "Jon",
            "last_name": "Doe",
            "zip": "12345",
            "password": "Password
        }
}
```

You'll note that nested in the `user` key is a hash containing `email`.  It also contains:

```
"first_name"
"last_name"
"zip"
"password"
```

You'll want to add those other attributes to the strong parameters, so they are available as well.

PixelWitch: done.

hpjaj:
@kylemh - Merging this.  How does the frontend PR for it look?

PR #200, opened 2017-11-24
Revert "Added an entrypoint to docker-compose to run database migrations when…"

hpjaj:
Reverts OperationCode/operationcode_backend#186

Caused the Code Schools API to go down.  

From @ohaiwalt during the outage:

```
operationcode-backend-3158010337-rjhfz           1/2       CrashLoopBackOff   320        1d
operationcode-backend-3158010337-zj8z0           1/2       CrashLoopBackOff   320        1d
```

Will need to find another solution for running database migrations.


hpjaj:
@ohaiwalt - Any reason we do **not** want to merge this revert?

PR #211, opened 2017-12-16
Feature/210 email only sign up endpoint

### LINKED ISSUE #210: API Endpoint for adding email address to SendGrid mailing list.
# Feature
Our business growth person has request that we add an email only sign up form to the front page of the website.  This allows someone to sign up for our email list (which already exists and is kept in sendgrid), without having to become a full member of Operation Code.

This feature is split between the front end and back end projects.  Front End is responsible for adding the form itself - and wiring it up to call a backend API call.  [This ticket captures the details for the front end](https://github.com/OperationCode/operationcode_frontend/issues/745).

The backend needs to 

- [x] Add an API endpoint - which takes an email address and sends it to the existing Sendgrid email list (When a member joins Operation Code, they are added to that same email list - so we can likely leverage existing code).  I don't think we need to add any secrets for auth to Sendgrid - as those likely already exist and are being used in the existing integration.

hpjaj:
# Description of changes
We want to allow someone to sign up for our email list, without having to become a full member of Operation Code (i.e. a "guest").  

This PR:
- [x] adds an API endpoint, which takes a guest's email address, and sends it to the existing Sendgrid email list
- [x] adds test coverage for new functionality 
- [x] updates our Apiary API docs with the new endpoint
- [x] as a citizen of the codebase, makes some styling updates for increased readability

# Issue Resolved

Fixes #210 

### Associated frontend issue

https://github.com/OperationCode/operationcode_frontend/issues/745

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -190,6 +190,40 @@ API endpoints that Operation Code's Rails backend makes available to its React f
             errors: "Some error message"
         }
 
+## Email List Recipients | Create [/api/v1/email_list_recipients{?email}]
++ Parameters
+
+  + email (string, required) - The email address of the OC guest to be added to the OC's email list
+
+### Add a guest (non-user) to the OC's email list [POST]
+
+  + Request (application/json)
+
+      + Headers
+
+              Authorization: Bearer Access-Token
+
+      + Body
+
+                  {
+                      "email": "john@gmail.com",
+                  }
+
+  + Response 201 (application/json)
+
+      + Body
+
+              {
+                  "email": "john@gmail.com",
+                  "guest": true
+              }
+
+  + Response 422 (application/json)
+
+          {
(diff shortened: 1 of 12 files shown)


hpjaj:
@nellshamrell @leenyburger - A "guest" cannot be an instance of the `User` class. As such, I created the guest as a `Struct` so that we can leverage the existing [SendGridClient class](https://github.com/OperationCode/operationcode_backend/blob/master/app/lib/send_grid_client.rb), without any custom refactoring.

You'll note that the class expects a `user` object with the following methods available on it:

```ruby
user.id
user.first_name
user.last_name
user.email
user.attributes
```

IMO, a `Struct` fits nicely in this use case.

nellshamrell:
Feel free to merge at will :)

PR #228, opened 2017-12-30
bug/addresses flakey controller test

### LINKED ISSUE #227: Bug/flaky controller test
# Bug Report
A test in the [locations_controller_test](https://github.com/OperationCode/operationcode_backend/blob/master/test/controllers/api/v1/locations_controller_test.rb#L110) randomly fails, causing [random build failures](https://travis-ci.org/OperationCode/operationcode_backend/builds/322703811?utm_source=github_status&utm_medium=notification).

This has happened in a pull request, as well as a master build on a green (different) pull request.

This test should consistently pass.

hpjaj:
# Description of changes

As described in issue #227 , a test in the `locations_controller_test` randomly fails, causing random build failures. 

This was happening because _sometimes_ there would be a `Location` record in the database equal to the hardcoded id of `1`.  When this happened, the record was being successfully deleted, thusly yielding a status of `200`, instead of the expected `422`.

By dynamically grabbing the `Location`s id, and adding 1000 to it, we ensure the passed location ID is not in the test db.

This also addresses a database issue in the local dev server logs, where an `.env` variable needed to be set.

# Issue Resolved

Fixes #227

### DIFF
--- a/.env
+++ b/.env
@@ -1,3 +1,4 @@
 COMPOSE_PROJECT_NAME=operationcodebackend
 REDIS_URL=redis://redis:6379/0
 RAILS_ENV=development
+POSTGRES_HOST=operationcode-psql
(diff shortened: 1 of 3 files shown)


hpjaj:
@nellshamrell @hollomancer @kylemh - Can I get a quick approval on this.  Gracias

PR #229, opened 2018-01-04
Issue #218 Added Requests API documentation

### LINKED ISSUE #218: Add API Documentation for Requests endpoints
# Feature

Our contributors need to be able to research, confirm and build against the API endpoints that the backend offers.

Document the following backend API endpoints in our [online API documentation](https://operationcodeapi.docs.apiary.io/#).  When doing so, **please follow** the steps described in our `README` for [how to document our API](https://github.com/OperationCode/operationcode_backend#api-documentation).

The endpoints to be documented are:

- [ ] `api/v1/requests_controller#index`
- [ ] `api/v1/requests_controller#create`
- [ ] `api/v1/requests_controller#show`
- [ ] `api/v1/requests_controller#update`

### LINKED ISSUE #219: Add API Documentation for Scholarships endpoints
# Feature

Our contributors need to be able to research, confirm and build against the API endpoints that the backend offers.

Document the following backend API endpoints in our [online API documentation](https://operationcodeapi.docs.apiary.io/#).  When doing so, **please follow** the steps described in our `README` for [how to document our API](https://github.com/OperationCode/operationcode_backend#api-documentation).

The endpoints to be documented are:

- [ ] `api/v1/scholarships_controller#index`
- [ ] `api/v1/scholarships_controller#show`
- [ ] `api/v1/scholarship_applications_controller#create`

wimo7083:
# Description of changes
Completes request for Issue #218  adds documentation for:
 api/v1/requests_controller#index
 api/v1/requests_controller#create
 api/v1/requests_controller#show
 api/v1/requests_controller#update

# Issue Resolved

Fixes #218

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -662,9 +662,10 @@ API endpoints that Operation Code's Rails backend makes available to its React f
             error: 'Unauthorized'
         }
 
-## Request | Update [/api/v1/requests{user_id}{?details,language,service_id,requested_mentor_id,status,assigned_mentor_id}]
+## Request | Update [/api/v1/requests/{?user_id,id,details,language,service_id,requested_mentor_id,status,assigned_mentor_id}]
 + Parameters
     
+    + id (integer, required) - Identification of the Mentor Request in the form on an integer
     + user_id (integer, required) - Identification of the User requesting a mentor in the form of an integer
     + details (string, optional) - Details of the mentor request in the form of a string
     + language (string, optional) - Primary language of the requesting mentee in the form of a string
@@ -685,6 +686,7 @@ API endpoints that Operation Code's Rails backend makes available to its React f
 
             {
                 "user_id": 5200,
+                "id": 4521,
                 "details": "I want to program",
                 "language": "English",
                 "service_id": 210,
@@ -698,8 +700,8 @@ API endpoints that Operation Code's Rails backend makes available to its React f
     + Body
 
             {
-                "id":4521,
                 "user_id": 5200,
+                "id":4521,
                 "details": "I want to program",
                 "language": "English",
                 "service_id": 210,
--- a/apiary.apib
+++ b/apiary.apib
@@ -542,9 +542,10 @@ API endpoints that Operation Code's Rails backend makes available to its React f
             errors: "Some error message"
         }
 
-## Request | Request a Mentor [/api/v1/requests{?user_id,details,language,service_id,requested_mentor_id,status,assigned_mentor_id}]
+## Request | Request a Mentor [/api/v1/requests{?id,user_id,details,language,service_id,requested_mentor_id,status,assigned_mentor_id}]
 + Parameters
     
+    + id (integer, required) - ID of mentor request in the form of an integer
     + user_id (integer, required) - ID of the user requesting a mentor, the passed in user must be a verified military veteran( i.e. user.verified == true)
     + details (string, optional) - Details of the mentor request in the form of a string
     + language (string, optional) - Primary language of the requesting mentee in the form of a string
@@ -563,7 +564,8 @@ API endpoints that Operation Code's Rails backend makes available to its React f
 
     + Body
 
-            {
+            {   
+                "id": 2130,
                 "user_id": 4600,
                 "details": "I want to type well",
                 "language": "English",
@@ -593,7 +595,7 @@ API endpoints that Operation Code's Rails backend makes available to its React f
 ## Request | Show a Mentor Request [/api/v1/requests/{id}]
 + Parameters
     
-    + id (integer, required) - Indentification of mentor request in the form of an integer
+    + id (integer, required) - Identification of mentor request in the form of an integer
 
     
 ### Show an Exiting Request From User ID [GET]
--- a/apiary.apib
+++ b/apiary.apib
@@ -645,7 +645,7 @@ API endpoints that Operation Code's Rails backend makes available to its React f
 
         [
             {   
-                "id": 251",
+                "id": 251,
                 "service_id": 223,
                 "language": "English",
                 "details": "I want to stop eating crayons",
--- a/apiary.apib
+++ b/apiary.apib
@@ -542,10 +542,10 @@ API endpoints that Operation Code's Rails backend makes available to its React f
             errors: "Some error message"
         }
 
-## Mentor Request | Create a New Mentor Request [/api/v1/requests{?user_id,details,language,service_id,requested_mentor_id,status,assigned_mentor_id}]
+## Request | Request a Mentor [/api/v1/requests{?user_id,details,language,service_id,requested_mentor_id,status,assigned_mentor_id}]
 + Parameters
     
-    + user_id (integer, required) - Identification of the User requesting a mentor in the form of an integer
+    + user_id (integer, required) - ID of the user requesting a mentor, the passed in user must be a verified military veteran( i.e. user.verified == true)
     + details (string, optional) - Details of the mentor request in the form of a string
     + language (string, optional) - Primary language of the requesting mentee in the form of a string
     + service_id (integer, optional) - Indentification of the mentorship request in the form of an integer
@@ -588,15 +588,9 @@ API endpoints that Operation Code's Rails backend makes available to its React f
                 "assigned_mentor_id": 250
             }
 
-+ Response 422 (application/json)
-
-    + Body
-
-            {
-                errors: "Some error message"
-            }
++ Response 401 (application/json)
   
-## Mentor Request | Show [/api/v1/requests/{request_id}]
+## Request | Show a Mentor Request [/api/v1/requests/{request_id}]
 + Parameters
     
     + request_id (integer, required) - Indentification of mentor request in the form of an integer
@@ -632,7 +626,7 @@ API endpoints that Operation Code's Rails backend makes available to its React f
                 "assigned_mentor_id": 260
             }
 
-## Mentor Requests | Unclaimed Collection [/api/v1/requests]
+## Request | Collection [/api/v1/requests]
 
 ### Get All Unclaimed Mentor Requests [GET]
 + Request (application/json)
@@ -675,7 +669,7 @@ API endpoints that Operation Code's Rails backend makes available to its React f
             error: 'Unauthorized'
         }
 
-## Mentor Request | Update [/api/v1/requests{user_id}{?details,language,service_id,requested_mentor_id,status,assigned_mentor_id}]
+## Request | Update [/api/v1/requests{user_id}{?details,language,service_id,requested_mentor_id,status,assigned_mentor_id}]
 + Parameters
     
     + user_id (integer, required) - Identification of the User requesting a mentor in the form of an integer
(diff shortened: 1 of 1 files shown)


apiary.apib
```diff
@@ -542,7 +542,183 @@ API endpoints that Operation Code's Rails backend makes available to its React f
             errors: "Some error message"
         }
 
+## Mentor Request | Create [/api/v1/requests{?user_id,details,language,service_id,requested_mentor_id,status,assigned_mentor_id}]
```
hpjaj: Since the model name is `Request` and the endpoint is also `requests`, can you please change this to "Request a Mentor", and then re-alphabetize all of the `Request` related endpoints accordingly.

apiary.apib
```diff
@@ -542,7 +542,183 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
+                "details": "I want to program",
+                "language": "English",
+                "service_id": 210,
+                "requested_mentor_id": 260,
+                "status": "Closed",
+                "assigned_mentor_id": 260
+            }
+
+
+## Mentor Request | Update [/api/v1/requests{?user_id,details,language,service_id,requested_mentor_id,status,assigned_mentor_id}]
```
hpjaj: The `update` endpoint takes the `request_id` as a part of the url itself, not a parameter.

Here is an example from [TeamMember#update](https://github.com/OperationCode/operationcode_backend/blob/master/apiary.apib#L896).

apiary.apib
```diff
@@ -542,7 +542,183 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
+                "status": "Open",
+                "assigned_mentor_id": 250
+            }
+
++ Response 201 (application/json)
+
+    + Body
+
+            {
+                "user_id": 4600,
```
hpjaj: This will also include the `id: ` of the newly created `Request` db record.

apiary.apib
```diff
@@ -542,7 +542,183 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
+
++ Response 422 (application/json)
+
+    + Body
+
+            {
+                errors: "Some error message"
+            }
+
+## Mentor Request | Show [/api/v1/requests{?user_id}]
```
hpjaj: The `show` endpoint takes the `request_id` as a part of the url itself, not a parameter.

Here is an example from [CodeSchool#show](https://github.com/OperationCode/operationcode_backend/blob/master/apiary.apib#L90).

apiary.apib
```diff
@@ -542,7 +542,183 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
+                "language": "English",
+                "service_id": 210,
+                "requested_mentor_id": 260,
+                "status": "Closed",
+                "assigned_mentor_id": 260
+            }
+
++ Response 400 (application/json)
+
+## Mentor Requests Unclaimed | Show [/api/v1/requests]
```
hpjaj: This is actually "Collection" instead of a "Show"

apiary.apib
```diff
@@ -542,7 +542,183 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
+
+    + Headers
+
+            Authorization: Bearer Access-Token
+        
++ Response 200 (application/json)
+
+        [
+            {
+                "service_id": 223,
```
hpjaj: The `id:` will be included for each one of these records

apiary.apib
```diff
@@ -542,7 +542,183 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
+                "status": "Closed",
+                "assigned_mentor_id": 260
+            }
+
++ Response 200 (application/json)
+
+    + Body
+
+            {
+                "user_id": 5200,
```
hpjaj: The `id:` will be included for the record

apiary.apib
```diff
@@ -749,6 +925,98 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
+
++ Response 404 (application/json)
+
+        {
+            error: 'No such record'
+        }
+
+## Scholarship Applications | Create [/api/v1/code_schools/{?scholarship_application,reason,terms_accepted,scholarship_id}]
++ Parameters
+
```
hpjaj: `user_id` is a required param

apiary.apib
```diff
@@ -749,6 +925,98 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
+            "updated_at": "2017-06-30T15:05:39.242Z"
+        }
+
++ Response 404 (application/json)
+
+        {
+            error: 'No such record'
+        }
+
+## Scholarship Applications | Create [/api/v1/code_schools/{?scholarship_application,reason,terms_accepted,scholarship_id}]
```
hpjaj: The URL is `/api/v1/scholarship_applications`

hpjaj:
@wimo7083 - Can you please pull out the code changes for Issue #218 out of this PR, and create a dedicated PR for that issue.  So that we have a 1:1 relationship, in that every issue has its own resolving PR.

wimo7083:
Updated, split PR, and renamed since you reviewed mentorships in this one.

apiary.apib
```diff
@@ -542,7 +542,186 @@ API endpoints that Operation Code's Rails backend makes available to its React f
             errors: "Some error message"
         }
 
+## Mentor Request | Create a New Mentor Request [/api/v1/requests{?user_id,details,language,service_id,requested_mentor_id,status,assigned_mentor_id}]
```
hpjaj: Since the model name is `Request` and the endpoint is also `../requests`, can you please all of these endpoints to begin with "Request ....".  For example, "Request a Mentor". 

And then re-alphabetize all of the Request related endpoints accordingly.

apiary.apib
```diff
@@ -542,7 +542,186 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
++ Parameters
+    
+    + user_id (integer, required) - Identification of the User requesting a mentor in the form of an integer
```
hpjaj: Since the passed `user` must also be verified, perhaps also include that information in this description.  i.e. "The passed in user must be a verified military veteran (i.e. `user.verified == true`)."

apiary.apib
```diff
@@ -542,7 +542,186 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
+                "user_id": 4600,
+                "details": "I want to type well",
+                "language": "English",
+                "service_id": 200,
+                "requested_mentor_id": 250,
+                "status": "Open",
+                "assigned_mentor_id": 250
+            }
+
++ Response 422 (application/json)
```
hpjaj: The [create action](https://github.com/OperationCode/operationcode_backend/blob/master/app/controllers/api/v1/requests_controller.rb#L11-L14) is current **not** accounting for an error response.  

Here is [an example](https://github.com/OperationCode/operationcode_backend/blob/master/app/controllers/api/v1/team_members_controller.rb#L15) of a different controller method that **is** accounting for error handling.


As such, you can adjust this whole section to return a `401`, like you did in the `index` action.

apiary.apib
```diff
@@ -542,7 +542,186 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
+                "request_id": 1234
+
+            }
+
++ Response 200 (application/json)
+
+    + Body
+
+            {
+                "request_id": 1234,
```
hpjaj: This attribute will just be `"id"`

apiary.apib
```diff
@@ -542,7 +542,186 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
+                errors: "Some error message"
+            }
+  
+## Mentor Request | Show [/api/v1/requests/{request_id}]
++ Parameters
+    
+    + request_id (integer, required) - Indentification of mentor request in the form of an integer
+
+    
+### Show an Exiting Request From User ID [POST]
```
hpjaj: This is a `get` request

apiary.apib
```diff
@@ -542,7 +542,186 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
+
++ Request (application/json)
+
+    + Headers
+
+            Authorization: Bearer Access-Token
+
+    + Body
```
hpjaj: In this case, the `id` is supplied as part of the `url` itself, and there will be no body in the request.

apiary.apib
```diff
@@ -542,7 +542,186 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
+            }
+        ]
+        
++ Response 401 (application/json)
+
+        {
+            error: 'Unauthorized'
+        }
+
+## Mentor Request | Update [/api/v1/requests{user_id}{?details,language,service_id,requested_mentor_id,status,assigned_mentor_id}]
```
hpjaj: There will be a `/` after requests because the `id` is part of the url itself, i.e. `/api/v1/requests/{user_id}{?details,language....`

apiary.apib
wimo7083: I'll fix them to be like this: 
`## Mentor Request | Request .... r [/api/vi/....]`

hpjaj: They'll need to go by their **model**, `Request`, so it would be:

```
## Request | Request a Mentor .... r [/api/vi/....]
```

apiary.apib
wimo7083: I'm confused, so there's no body in the update of the request?

[Frontend](https://github.com/OperationCode/operationcode_frontend/blob/e2de0bc04de963aa8181ea2d7bae60f42460597e/src/shared/utils/apiHelper.js#L56), to me this looks like it's passing the id in the url like you say, but there's a body in the request that contains `status` and `mentor`. But then [here](https://github.com/wimo7083/operationcode_backend/blob/01e2a2d4c8fc91a869545ede3e32a1114bced5c5/app/controllers/api/v1/requests_controller.rb#L33) shows that the update takes in request params. 

I compared the syntax for both update with [Request](https://github.com/wimo7083/operationcode_backend/blob/master/app/controllers/api/v1/requests_controller.rb) controller and [User](https://github.com/wimo7083/operationcode_backend/blob/master/app/controllers/api/v1/users_controller.rb) controller and they both have the same type of input with params. The only difference is one uses `obj.update` and the other uses `obj.update_attributes`.

So i thought maybe this has to do with the difference. But [this](https://stackoverflow.com/a/27684300) states that the only difference in behavior of the two functions it the return value for a success vs failure. `update_attributes` returns false if it fails validation, `update` returns the new object if it fails validation. 

At this point I'm thorougly confused why there isn't a body in the request. 

I also used a `PUT` here but changed it to `PATCH` based on the information provided in the frontend axios call.

apiary.apib
wimo7083: I included the value, but I think because I did `"id"` and not `"id:"` github isn't registering correctly?

apiary.apib
hpjaj: @wimo7083 - Apologies, I miscommunicated that.  My intent was to say that the `id` would not be in the body, rather in the url itself.  And yes, you're correct, there is a body in the request with all the attributes you have below.

apiary.apib
hpjaj: Yeah, force of (Ruby) habit.  In this API documentation it is `"id"`.

PR #233, opened 2018-01-11
force sidekiq to fallback to config/database.yml credentials

tskuse:
# Description of changes
This PR prevents the defining of the environment variable `DATABASE_URL`. Leaving `DATABASE_URL` undefined forces Sidekiq to fallback to using `config/database.yml` for the Postgres configuration. Without this change, `DATABASE_URL` evaluates to an invalid URL format and Sidekiq attempts to connect to a database named `operationcode-psql` which does not exist.

# Issue Resolved
Sidekiq database connection failure


hpjaj:
@ohaiwalt - Can you pls approve or comment on this PR.  Thanks!

ohaiwalt:
Do we know that Sidekiq will pick up other environments that aren't default? 

Also, prefer delete over commenting.

PR #232, opened 2018-01-09
Included config for sidekiq

wimo7083:
# Description of changes

Added config file for sidekiq, and adjusted config reference

# Issue Resolved

Fixes `docker-compose up --build` and that error related to sidekiq throwing many exceptions.

PR #234, opened 2018-01-11
Revert "Included config for sidekiq"

hpjaj:
Reverts OperationCode/operationcode_backend#232


robbkidd:
Any recollection on why this change needed to be reverted? A sidekiq config to set queue names and configure concurrency and timeouts would be good.

I can guess that maybe the omission of `.erb` (Embedded RuBy) from the file name means that the Ruby that was embedded in the YAML—`<%= hostname.strip %>`—did not get interpreted and Sidekiq was like "Whaaa?" trying to use that as a queue name.

PR #236, opened 2018-01-11
Add is_mooc to CodeSchools and create get_moocs endpoint

### LINKED ISSUE #169: Support MOOCs on the code school resource/model/table
# Feature
[This issue](https://github.com/OperationCode/operationcode_frontend/issues/381) on the front-end calls for MOOCs to stop being hardcoded and be returned as JSON from the API like most, if not all, of everything else.

Per some conversation on the front-end issue, we think the existing [code school](https://github.com/OperationCode/operationcode_backend/blob/master/app/models/code_school.rb) model can support MOOCs by adding an `isMooc` boolean via a migration.

This is a multi-step feature request, please let me know if each of the two parts should be seperated into an issue by themselves. I do not see any written tests that cover the code school model/resource, please let me know if I've overlooked those and a third bullet point needs added that covers writing tests for the new column.

* Add a migration to add `isMooc` as a boolean to the code school table
    * Make sure the code school model is aware of this new column
* Add a method to the code school controller that finds all code schools where `isMooc` is true and returns JSON

I am comfortable tackling this if those in charge of op code back-end are comfortable with it. I have some RoR experience but it has been several years since I worked in an active Rails codebase. I also plan on doing the front-end pull request once this is complete.

leenyburger:
# Description of changes
This PR adds a boolean is_mooc to the CodeSchools table and creates a /get_moocs endpoint for the front end. 
- is_mooc boolean is added to the CodeSchools table via a migration
- get_moocs action is added to the code_schools_controller to return all CodeSchools with is_mooc set to true
- Tests added to code_schools_controller_test.rb
- Route added to access endpoint
- Apiary documentation updated to reflect new endpoint

## QA
[ ] In the console, add several Code Schools. Give each school a different name and set is_mooc to true on some and false on others: 
```  
school1 = CodeSchool.create(name: "MySchools", url: "https://www.school.com", logo: "Mylogo", full_time:"false", hardware_included: "true", has_online: "true", online_only:"false", is_mooc:"false")
```
[ ] Running server, go to /api/v1/code_schools/get_moocs
[ ] You should see the Code Schools created ^ with is_mooc set to true, and not the ones with is_mooc set to false. 

# Issue Resolved

Fixes #169

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -19,6 +19,7 @@ API endpoints that Operation Code's Rails backend makes available to its React f
                 "full_time": true,
                 "hardware_included": false,
                 "has_online": false,
+                "is_mooc": false,
                 "online_only": false,
                 "locations":  [
                       {
@@ -39,7 +40,42 @@ API endpoints that Operation Code's Rails backend makes available to its React f
             },
         ]
 
-## CodeSchool | Create [/api/v1/code_schools{?name,url,logo,full_time,hardware_included,has_online,online_only,notes}]
+## MOOC CodeSchool | Collection [/api/v1/code_schools/get_moocs]
+
+### List all MOOCS [GET]
+
++ Response 200 (application/json)
+
+        [
+            {
+                "name": "Galvanize",
+                "url": "https://new.galvanize.com", "logo": "https://raw.githubusercontent.com/OperationCode/operationcode_frontend/master/src/images/codeSchoolLogos/galvanize.jpg",
+                "logo": "https://raw.githubusercontent.com/OperationCode/operationcode_frontend/master/src/images/codeSchoolLogos/DPL_logo.png",
+                "full_time": true,
+                "hardware_included": false,
+                "has_online": false,
+                "is_mooc": true,
+                "online_only": false,
+                "locations":  [
+                      {
+                          "va_accepted": false,
+                          "address1": "44 Tehama Street",
+                          "city": "San Francisco",
+                          "state": "CA",
+                          "zip": 94015
+                      },
+                      {
+                          "va_accepted": true,
+                          "address1": "1062 Delaware Street",
+                          "city": "Denver",
+                          "state": "CO",
+                          "zip": 80204
+                      },
+                  ]
+            },
+        ]
+
+## CodeSchool | Create [/api/v1/code_schools{?name,url,logo,full_time,hardware_included,has_online,online_only,notes,is_mooc}]
 + Parameters
 
     + name (string, required) - The CodeSchool's name
@@ -50,6 +86,7 @@ API endpoints that Operation Code's Rails backend makes available to its React f
     + has_online (boolean, required) - Does the CodeSchool have an online component
     + online_only (boolean, required) - Is the CodeSchool online only?
     + notes (text, optional) - Any additional information about the CodeSchool
+    + is_mooc (boolean, optional) - Is this a MOOC?
 
 ### Create a CodeSchool Member [POST]
 
@@ -69,7 +106,8 @@ API endpoints that Operation Code's Rails backend makes available to its React f
                       "full_time": "false",
                       "hardware_included": "false",
                       "has_online": "true",
-                      "online_only": "false"
+                      "online_only": "false",
+                      "is_mooc": "false"
                       }
                   }
 
@@ -108,6 +146,7 @@ API endpoints that Operation Code's Rails backend makes available to its React f
             "created_at": "2017-09-26T06:54:58.607Z",
             "updated_at": "2017-09-26T06:54:58.607Z",
             "notes": null
+            "is_mooc": "false"
         }
 
 + Response 404 (application/json)
@@ -146,7 +185,7 @@ API endpoints that Operation Code's Rails backend makes available to its React f
                   errors: "Some error message"
               }
 
-## CodeSchool | Update [/api/v1/code_schools/{code_school_id}{?name,url,logo,full_time,hardware_included,has_online,online_only,notes}]
+## CodeSchool | Update [/api/v1/code_schools/{code_school_id}{?name,url,logo,full_time,hardware_included,has_online,online_only,notes,is_mooc}]
 
 + Parameters
 
@@ -158,6 +197,7 @@ API endpoints that Operation Code's Rails backend makes available to its React f
     + has_online (boolean, required) - Does the CodeSchool have an online component
     + online_only (boolean, required) - Is the CodeSchool online only?
     + notes (text, optional) - Any additional information about the CodeSchool
+    + is_mooc (boolean, optional) - Is the school a MOOC?
 
 ### Update an Existing CodeSchool [PUT]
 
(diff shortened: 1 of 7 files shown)


app/controllers/api/v1/code_schools_controller.rb
```diff
@@ -42,10 +42,19 @@ def destroy
         end
       end
 
+      def get_moocs
+        mooc_school = CodeSchool.where(is_mooc: "true")
+        if mooc_school
```
hpjaj: This conditional will return false positives.  If the ActiveRecord collection comes back empty, it will still satisfy the check, and would render an empty set of data.  For example:

```ruby
[13] pry(main)> schools = CodeSchool.where(id: 99999)
=> []
[11] pry(main)> if schools
[11] pry(main)*   p 'something'
[11] pry(main)* else
[11] pry(main)*   p 'nada'
[11] pry(main)* end
"something"
=> "something"
```

To account for this, one option would be to use the `.present?` method:

```ruby
[14] pry(main)> schools.present?
=> false
```

That all said, since this is a straight `GET` without any params, I think in this case we can safely just render the results, and remove the conditional completely.  That way, if there are results, they will be rendered.  And if there are no results, then that will be reflecting in the rendering, as well.

And there's really no error handling you would need to account for here.

app/controllers/api/v1/code_schools_controller.rb
```diff
@@ -42,10 +42,19 @@ def destroy
         end
       end
 
+      def get_moocs
```
hpjaj: Can we change the name of this action to be in keeping the `scope` like nature of its purpose.  And just call it `moocs`.

This way the url endpoint can also read like a scope on the model, i.e. `api/v1/code_schools/moocs`.

You'll want to adjust the route, accordingly.

app/controllers/api/v1/code_schools_controller.rb
```diff
@@ -42,10 +42,19 @@ def destroy
...
+        mooc_school = CodeSchool.where(is_mooc: "true")
```
hpjaj: A couple things here.

1 - Small nit, can we change the variable from singular to plural `mooc_schools`
2 - `is_mooc` is a `boolean` type, not string. So this would actually be `CodeSchool.where(is_mooc: true)`

test/controllers/api/v1/code_schools_controller_test.rb
```diff
@@ -58,4 +58,18 @@ class Api::V1::CodeSchoolsControllerTest < ActionDispatch::IntegrationTest
     get api_v1_code_school_path(id), as: :json
     assert_response :missing
   end
+
+  test ":get_moocs returns schools with is_Mooc set to true" do
+    @school.is_mooc = "true"
```
hpjaj: You'll want to update this strings to booleans, i.e. `true` and `false`

apiary.apib
```diff
@@ -39,7 +40,42 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
         ]
 
-## CodeSchool | Create [/api/v1/code_schools{?name,url,logo,full_time,hardware_included,has_online,online_only,notes}]
+## MOOC CodeSchool | Collection [/api/v1/code_schools/get_moocs]
+
+### List all MOOCS [GET]
+
++ Response 200 (application/json)
+
+        [
```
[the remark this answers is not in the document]
leenyburger: Good question. Do you know who is working this on the front end? I checked out the linked issue, but looks like there hasn't been chatter since Oct.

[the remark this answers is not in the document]
hpjaj: @leenyburger - Let's have this be in keeping with the structure that the current [CodeSchools - Collection](https://operationcodeapi.docs.apiary.io/#reference/0/codeschool-collection/list-all-codeschool-members) endpoint returns.

cc: @kylemh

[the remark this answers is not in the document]
kylemh: It's a low priority issue to be sure since we don't add MOOCs often.

I haven't seen this issue on the back-end, but I would prefer that MOOCs not need to be a separate API call. Instead have an `is_mooc` boolean for all existing code schools should be sufficient.

kylemh:
Aside from my in-review comment, we could also make this very easy by simply including MOOCs into the backend with `null` address field information. When filtering code school JSON on the front-end, we can look for `school.online_only && school.address1 && school.city && school.state && school.zip` - with that filtration, MOOCs should be the only results.

kylemh:
However, maybe there is some benefit to modeling MOOCs in the back-end for some future feature/use case

hpjaj:
@kylemh - What if the frontend just filters the JSON from the existing [Code Schools collection endpoint](https://operationcodeapi.docs.apiary.io/#reference/0/codeschool-collection/list-all-codeschool-members), based on the new attribute?

For example, if this is the JSON that the frontend receives from the existing endpoint:

```
[
    {
        "name": "Galvanize",
        "url": "https://new.galvanize.com", "logo": "https://raw.githubusercontent.com/OperationCode/operationcode_frontend/master/src/images/codeSchoolLogos/galvanize.jpg",
        "logo": "https://raw.githubusercontent.com/OperationCode/operationcode_frontend/master/src/images/codeSchoolLogos/DPL_logo.png",
        "full_time": true,
        "hardware_included": false,
        "has_online": false,
        "online_only": false,
        "mooc": false,
        "locations":  [
              {
                  "va_accepted": false,
                  "address1": "44 Tehama Street",
                  "city": "San Francisco",
                  "state": "CA",
                  "zip": 94015
              },
              {
                  "va_accepted": true,
                  "address1": "1062 Delaware Street",
                  "city": "Denver",
                  "state": "CO",
                  "zip": 80204
              },
          ]
    },
]
```

You could just filter that JSON by the `mooc` boolean attribute.

No?

If that does not work, another option would be to pass optional parameters to the existing [Code Schools collection endpoint](https://operationcodeapi.docs.apiary.io/#reference/0/codeschool-collection/list-all-codeschool-members).  For example, optionally passing a param of `mooc: true`.

Thoughts?

kylemh:
@hpjaj hahah @leenyburger had the same idea - we discussed on Slack instead of GH. I agree!

hpjaj:
@kylemh - Ha...nice!

@leenyburger - This will allow you to trim a lot from this PR.  Can we please change the name of the new db attribute to either `mooc` or `offers_mooc`.

Let me know if there is anything I can help with!

leenyburger:
@hpjaj I think it will be easier to close this PR and create a new one instead of rolling back my commits. Let me know if you disagree.

hpjaj:
@leenyburger - That works

JennWeideman:
whats the current status of this I asking because this is referenced in frontend issue

hpjaj:
@JennWeideman - @leenyburger is currently working on a solution for it in https://github.com/OperationCode/operationcode_backend/pull/250

JennWeideman:
@hpjaj Thank you for getting back to me! Please ping me when we have an update

PR #240, opened 2018-01-15
feature/216 Endpoint for slack bot to interact with backend

### LINKED ISSUE #216: Request add endpoint to get user's verified state.
# Feature: New Apiary Endpoint to request user information
Currently [pybot](https://github.com/OperationCode/operation_code_pybot) provides new slack members with useful resources. per front end issue [#765](https://github.com/OperationCode/operationcode_frontend/issues/765 ). Only members who are verified veterans are eligible for certain resources ex: mentorship, and veterans scholarships. We shouldn't offer these resources to individuals who don't qualify. 

I'd like to be able to get an endpoint in which I can take a registered user information which is connected to a slack id to user information. Additionally there are other sections within the [User Table](https://github.com/OperationCode/operationcode_backend/blob/master/db/schema.rb#L184-L225) that would be helpful if exposed (mentor, details about service, interests and contact information). 

Since this is being requested by an external program some of these details might be important to keep secure. I'm unsure if the API is public facing or if we need to make a way for only registered applications/ requests to get the information.

wimo7083:
# Endpoint for slack bot

Currently working on trying to get the slack bot to interact with the front end this is currently not working but trying to raise discussion and get help.

Things that need to be done:

- [x] api documentation
- [ ] making the connection secured so not anybody can get data
- [x] registering the `https://operationcode.org/api/v1/slack/users_info` endpoint
- [ ] discussion on how to implement better

# Issue Resolved attempt at endpoint addition

Fixes #216

The reason for this:
1. Get a list of all users to see who doesn't have a registered slack account, then add slack account using the slack api
2. Continually send slack user information to the database when we get new users.
3. Get single individual information (interests, slack_name, etc) using email as the key. 

I don't have a way to get the user_id, and the slack bot doesn't login, so I'm working on creating a unique token that both the database have and my slack bot has to verify before doing any of these requests. 

Right now my biggest issue is getting around not being able to implement the route or test locally. 

``` D, [2018-01-15T21:03:45.767201 #13] DEBUG -- : No route matches [GET] "/api/v1/slack/user_info" excluded from capture: DSN not set
web_1                 | F, [2018-01-15T21:03:45.767623 #13] FATAL -- :
web_1                 | F, [2018-01-15T21:03:45.767663 #13] FATAL -- : ActionController::RoutingError (No route matches [GET] "/api/v1/slack/user_info"):
web_1                 | F, [2018-01-15T21:03:45.767677 #13] FATAL -- :
web_1                 | F, [2018-01-15T21:03:45.767694 #13] FATAL -- : actionpack (5.0.2) lib/action_dispatch/middleware/debug_exceptions.rb:53:in `call'```

### CODE REVIEW

File: app/controllers/api/v1/slack/user_info_controller.rb [hunk 1]
```diff
@@ -0,0 +1,44 @@
+module Api
+  module V1
+    module Slack
+      class UserInfoController < ApplicationController
+        before_action VerifyRequest 
+ 
+        # get all users 
+        def index
+          render json: Users.all
```
hpjaj: We probably do not need to send over every attribute on `User`.  We could instead update the [user serializer](https://github.com/OperationCode/operationcode_backend/blob/master/app/serializers/user_serializer.rb) and repurpose that.

app/controllers/api/v1/slack/user_info_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+            render json: user 
+          else
+            render json: { error: 'No such email record' }, status: :not_found
+          end
+        end
+
+        # update a single slack name by email
+        def update
+          user = User.find(params[:email])
+          if user.update(:slack_name params[:slack_name])
```
hpjaj: The syntax for this would be `if user.update(slack_name: params[:slack_name])`

app/controllers/api/v1/slack/user_info_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+            render json: user
```
hpjaj: We would render the serializer here, too

app/controllers/api/v1/slack/user_info_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+          end
+        end
+
+        # update a single slack name by email
+        def update
+          user = User.find(params[:email])
+          if user.update(:slack_name params[:slack_name])
+            render json: user 
+          else
+            render json: { errors: user.errors.full_messages }
```
hpjaj: We'd want to add in a `status: ...` here, too

app/controllers/api/v1/slack/user_info_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+          end
+        end
+      end 
+
```
hpjaj: We would add a `private` here, so that the `verify_request` method is not `public`.  [For example](https://github.com/OperationCode/operationcode_backend/blob/master/app/controllers/api/v1/resources_controller.rb#L51-L67).

app/controllers/api/v1/slack/user_info_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+          user = User.find(params[:email])
+          if user.update(:slack_name params[:slack_name])
+            render json: user 
+          else
+            render json: { errors: user.errors.full_messages }
+          end
+        end
+      end 
+      
+      class VerifyRequest
```
hpjaj: I haven't decided how we'd want to deal with this non-Devise request, but if we dedicate a class to it, that class would live outside of this file.  

So you can remove this class definition.

app/controllers/api/v1/slack/user_info_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+          def self.before(controller)
```
hpjaj: Regardless if the logic for the authentication ends up living in here or not, this would still be a private method.  i.e  `def verify_request`.

Instead of `controller`, the parameters for this method would be something that is passed in the `params` hash.  

And that parameter name could perhaps be `credentials`, or something like that.

app/controllers/api/v1/slack/user_info_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+            unless controller(:auth_key == ENV['pybot_token'] )
```
hpjaj: I need to noodle on the best way to deal with this one-off authentication.

app/controllers/api/v1/slack/user_info_controller.rb
wimo7083: Would the call be, I changed the serialization in the file but I'm not sure how to reference it. I thought that `Users.all` would use the default serialization. 
`render json: Users.attributes`

app/controllers/api/v1/slack/user_info_controller.rb
```diff
@@ -0,0 +1,44 @@
+module Api
+  module V1
+    module Slack
```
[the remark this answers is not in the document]
wimo7083: Okay, moved to the slack users controller.

hpjaj:
Closing this out in favor of the Slack integration issues 1 - 6

PR #245, opened 2018-01-29
TeamMember Groups Field and Board Members

### LINKED ISSUE #243: Add Team table group functionality
# Create a new column for Team database

Currently we treat all team members as general members, and when promotion/ demotion we have to remove from database and statically add them to front end. This change would make it so we can have different class of members for our team

We can then add legacy members, board members, and general members and allow the front end to decide how to render while correctly tracking their state instead of removing from database. 

Modify schema to include this column
Migrate database to new schema
Include board members who are not currently tracked.

alexbeeken:
# Description of changes
Adds groups field to TeamMember table.
Adds description to TeamMember table.
Adds image_src to TeamMember table.

Validates presence of :group in TeamMember model
Validates group is 'team', 'board' or 'executive'
Tests this in the TeamMember model

Adds data from frontend into db via reversible migration

Adds group, description and image_src to team_members_controller params.
Updates team_member_controller to account for those new params and for required group attribute.

Updates api docs to reflect these changes.

# Issue Resolved
Fixes #243


db/migrate/20180129223616_add_group_to_team_members.rb
```diff
@@ -0,0 +1,7 @@
+class AddGroupToTeamMembers < ActiveRecord::Migration[5.0]
+  def change
+    add_column :team_members, :description, :text
+    add_column :team_members, :group, :string, null: "false"
+    add_column :team_members, :image_src, :string, default: "images/img_unk.png"
```
hpjaj: Let's not set a default here for the time being, and allow for scoping this by `image_src: nil`.

db/migrate/20180129231309_add_board_members.rb
```diff
@@ -0,0 +1,131 @@
+class AddBoardMembers < ActiveRecord::Migration[5.0]
+  def up
+    TeamMember.create!(
```
hpjaj: Awesome job on assembling all of this.  A few thoughts:

1. Can we shift this over to a `rake` task, instead of a migration?  Preferring to make database alterations in a migration file, and data changes in a `rake` task.
2. In the `rake` task, can you check for the presence of the db record, before creating one?  And if you find one, update it.
3. In a separate `rake` task, can you update all of the other `TeamMember`s in the `/team` page with `group: "team"`

apiary.apib
```diff
@@ -1071,14 +1071,20 @@ API endpoints that Operation Code's Rails backend makes available to its React f
 
         [
             {
+                "description": "Worked for lots of successful companies.",
```
hpjaj: When I tested this in apiary's text editor, there are some _semantic_ issues.

Can you take a quick read through:  https://github.com/OperationCode/operationcode_backend#api-documentation

And see if you can resolve them?

test/factories/team_members.rb
```diff
@@ -2,5 +2,20 @@
   factory :team_member do
     name "John Smith"
     role "Board Advisor"
+    group "board"
+    description "Long and wonderful history of employment"
+    image_src "images/john_smith.jpg"
+  end
+
+  trait :board do
```
hpjaj: Nice.

Somewhat related to this, the `seeds.rb` file will need updating.  It is pulling its `TeamMember` data from:

https://github.com/OperationCode/operationcode_backend/blob/master/config/team_members.yml

Can you update that .yml file with `group` data, as this will now be a validated attribute.

db/migrate/20180129231309_add_board_members.rb
alexbeeken: Yes, there is already a similar rake task that I didn't notice before. I'll put it there.

alexbeeken: Or, rereading your comment, in a separate rake task.

alexbeeken: @hpjaj For now I'm going to move the board rake task into the seeds.rb script and remove the rake tasks for seeding boards and teams. It seems redundant to have both and I like having things in a yaml file.

When I finish up with this PR I think I'll do another branch and put that script into separate rake tasks with a :seed task for reseeding everything. That way a dev can run the whole thing or each one individually and we can keep all the data in .yamls.

test/factories/team_members.rb
alexbeeken: I updated the seeds.rb. Now it calls a new class to seed team members. I updated the rake tasks to use that same class and renamed/added tasks.

alexbeeken:
@hpjaj okay tackled all your questions and decided on a few things:
 1. I put the seeding into a single class (SeedTeamMembers) that can be called in both seeds and a new :seed_team_members rake task namespace.
2. API docs shouldn't have any more semantic problems
3. I put the yaml file paths and group names into constants so they can be easily changed.

Ready for another review! : D

app/lib/seed_team_members.rb
```diff
@@ -0,0 +1,33 @@
+class SeedTeamMembers
+  BOARD_DATA_PATH = 'board_members.yml'.freeze
+  TEAM_DATA_PATH = 'team_members.yml'.freeze
+
+  def self.from_yaml(file_name, group)
+    members_seed_file = Rails.root.join('config', file_name)
+    members = YAML::load_file(members_seed_file)
+    members.map do |member|
+      TeamMember.find_or_create_by!(
```
hpjaj: This could inadvertently create duplicate `TeamMember` records.  

The way it is written, it is trying to find a record that fits **all** of the `member` attributes, and if it does not find one, it creates a new one.

Instead we'd want it to find a `TeamMember` with the same `name`, and if it does not find one, then create it.  But if it does find one, it should **update** that record.  

The yaml file, and our codebase, does not have any duplicate names at this point, so this will work fine.  And this is a temporary solution until we have better access to prod's db.

One way to accomplish this is to pass a block to `find_or_create_by`, kind of like [this example from the docs](http://guides.rubyonrails.org/active_record_querying.html#find-or-create-by):

```ruby
Client.find_or_create_by(first_name: 'Andy') do |c|
  c.locked = false
end
```

app/lib/seed_team_members.rb
```diff
@@ -0,0 +1,33 @@
+class SeedTeamMembers
```
hpjaj: Love that you abstracted the logic for this into a dedicated class 👍

app/models/team_member.rb
```diff
@@ -1,3 +1,9 @@
 class TeamMember < ApplicationRecord
-  validates :name, :role, presence: true
+  BOARD_GROUP_NAME = "board".freeze
+  EXECUTIVE_GROUP_NAME = "executive".freeze
+  TEAM_GROUP_NAME = "team".freeze
+  validates :name, :role, :group, presence: true
```
hpjaj: Nit, pls add a space in between the group of constants and group of validations

app/models/team_member.rb
```diff
@@ -1,3 +1,9 @@
 class TeamMember < ApplicationRecord
-  validates :name, :role, presence: true
+  BOARD_GROUP_NAME = "board".freeze
```
hpjaj: 👍

alexbeeken:
@hpjaj I added some tests just to make sure this was behaving the way you were talking about. This should be ready for another review.

PR #181, opened 2017-10-14
Feature/175 admin dashboard

hpjaj:
# Description of changes

Creates an admin dashboard using [ActiveAdmin](https://activeadmin.info/index.html).

- [x] On the `User` page, it should provide a broad filtering mechanism to query for any number of one-off criteria, based on the `User` tables attributes
- [x] Can create, edit, view and/or delete `User`s
- [x] Can create, edit, view and/or delete `AdminUser`s

On the `User` page, it should be able to filter users by:
- [x] mentor
- [x] state
- [x] verified status
- [x] tag name

On the main dashboard, it should display
- [x] How many users today
- [x] How many users over last week
- [x] How many users over last month
- [x] How many users over last year
- [x] Number of verified users

On a `User`'s show page:
- [x] A button that tags/un-tags the user with "community leader"

# Issue Resolved

Fixes #175

PR #247, opened 2018-02-01
Feature/175 ActiveAdmin console

hpjaj:
Reverts OperationCode/operationcode_backend#246

Original PR https://github.com/OperationCode/operationcode_backend/pull/181

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,7 @@ git_source(:github) do |repo_name|
   "https://github.com/#{repo_name}.git"
 end
 
+gem 'activeadmin', '~> 1.1'
 gem 'active_model_serializers'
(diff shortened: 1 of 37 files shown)


hpjaj:
@nellshamrell - This commit: https://github.com/OperationCode/operationcode_backend/pull/247/commits/a6e76f81400bdf4102369e0505a67608afb7ba79 resolves the issue around a user being able to login.  

I was able to reproduce the issue we experienced when this was live in prod.  I have tested it locally, and this commit resolves things.

Now all we need to figure out is the issue with the ports, so that we can visit `https://operationcode.org/admin/login` in prod.

nellshamrell:
This looks good, feel free to deploy at will, I'm working on the infrastructure for this now :)

PR #248, opened 2018-02-02
Rely on Geocoder for validating zip codes

### LINKED ISSUE #242: Allow non-US zip codes for volunteers
Hello!

We have a developer in the U.K. who is interested in contributing to our open source projects.  He attempted to join our org as a volunteer, but we only accept U.S. zip codes.  We should consider accepting non-U.S. zip codes for volunteers.

alexbeeken:
# Description of changes
This PR adds an extra check before saving a new user record. This checks to see if the zip code is found by Geocoder by checking the `latitude` and `longitude` attributes for truthiness.

This is to allow any zip code that Geocoder can locate to be input during user before_create validations.

# Issue Resolved
Fixes #242


test/models/user_test.rb
```diff
@@ -46,6 +46,14 @@ def setup
...
   end
 
+  test 'accepts non-us zipcodes (UK)' do
+    u = build(:user, latitude: nil, longitude: nil)
+
+    u.update_attributes(zip: 'HP2 4HG')
+    assert_equal 40.7143528, u.latitude
+    assert_equal -74.0059731, u.longitude
+  end
+
```
hpjaj: Can you add a test for the negative case, where the new validation will not be satisfied, and the errors will be added

alexbeeken: So I fixed this this morning and it led me down a little rabbit hole of making all API requests mocked with VCR. This also involved replacing/removing some stubbing done with Geocoder and setting up a little Monkey patch I found. It creates a new cassette for every test automatically and names it the test name. It made all the stubs in user unnecessary save for one which I could remake using mocha.

alexbeeken:
Ready for review again. I added some tests for the negative case and revamped all api mocking to use VCR. This should make tests faster and not have to make any kind of actual API call after the first time they are run.

test/test_helper.rb
```diff
@@ -19,3 +19,19 @@ def authorization_headers(user)
   config.cassette_library_dir = 'test/cassettes'
   config.hook_into :webmock
 end
+
+class ActiveSupport::TestCase
+  def self.test(test_name, &block)
```
hpjaj: A couple things on this:

1.  Please add [this documentation](https://github.com/alexbeeken/operationcode_backend/blob/9bc9ce4b6d15aca62dcff7a52f486bacbcb30600/test/support/vcr.rb#L1-L12) to this, for future devs understanding.  Also, if you can find the url where you found this solution, please include that as an `@see ...`, like [this](https://github.com/OperationCode/operationcode_backend/blob/master/app/controllers/api/v1/users/passwords_controller.rb#L8).

2.  Please add this method to the [ActiveSupport::TestCase](https://github.com/OperationCode/operationcode_backend/blob/master/test/test_helper.rb#L8-L15) class.

app/models/user.rb
```diff
@@ -27,6 +27,7 @@ class User < ApplicationRecord
   before_save :geocode, if: ->(v) { v.zip.present? && v.zip_changed? }
   before_save :upcase_state
   before_save :downcase_email
+  before_save :validate_zipcode
```
hpjaj: Instead of a callback, can we convert this into a [custom validation method](http://guides.rubyonrails.org/active_record_validations.html#custom-methods).

app/models/user.rb
```diff
@@ -104,6 +105,12 @@ def token
     JsonWebToken.encode(user_id: self.id, roles: [], email: self.email, verified: verified)
   end
 
+  def validate_zipcode
```
hpjaj: As such, this would become a `private` method.

app/models/user.rb
```diff
@@ -104,6 +105,12 @@ def token
...
+    return true if longitude && latitude
+    errors['zip code'] << 'not found'
+    raise_validation_error
```
hpjaj: In that way, Rails would take care of this

app/models/user.rb
```diff
@@ -104,6 +105,12 @@ def token
     JsonWebToken.encode(user_id: self.id, roles: [], email: self.email, verified: verified)
   end
 
+  def validate_zipcode
+    return true if longitude && latitude
```
hpjaj: Then here you could just `return if ...` vs `return true if`...

app/models/user.rb
```diff
@@ -104,6 +105,12 @@ def token
...
+    errors['zip code'] << 'not found'
```
hpjaj: Here you would follow this pattern, for example:

```ruby
errors.add(:expiration_date, "can't be in the past")
```

hpjaj:
@alexbeeken - Are you able to make the above code review suggestions?  Hope so...this is a fantastic PR!

alexbeeken:
@hpjaj Hey sorry for the delay, you know how life gets busy. I updated based on your comments and this is ready for review again.

PR #250, opened 2018-02-05
Added boolean mooc to code schools table

### LINKED ISSUE #169: Support MOOCs on the code school resource/model/table

leenyburger:
# Description of changes
Add boolean "mooc" to code schools table. See closed PR #236 for discussion. 

# Issue Resolved

Fixes #169

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -20,6 +20,7 @@ API endpoints that Operation Code's Rails backend makes available to its React f
                 "hardware_included": false,
                 "has_online": false,
                 "online_only": false,
+                "mooc": false
                 "locations":  [
                       {
                           "va_accepted": false,
@@ -39,7 +40,7 @@ API endpoints that Operation Code's Rails backend makes available to its React f
             },
         ]
 
-## CodeSchool | Create [/api/v1/code_schools{?name,url,logo,full_time,hardware_included,has_online,online_only,notes}]
+## CodeSchool | Create [/api/v1/code_schools{?name,url,logo,full_time,hardware_included,has_online,online_only,notes,mooc}]
 + Parameters
 
     + name (string, required) - The CodeSchool's name
@@ -50,6 +51,7 @@ API endpoints that Operation Code's Rails backend makes available to its React f
     + has_online (boolean, required) - Does the CodeSchool have an online component
     + online_only (boolean, required) - Is the CodeSchool online only?
     + notes (text, optional) - Any additional information about the CodeSchool
+    + mooc (boolean, optional) - Does the CodeSchool offer Massive Open Online Courses?
 
 ### Create a CodeSchool Member [POST]
 
@@ -70,6 +72,7 @@ API endpoints that Operation Code's Rails backend makes available to its React f
                       "hardware_included": "false",
                       "has_online": "true",
                       "online_only": "false"
+                      "mooc": "false"
                       }
                   }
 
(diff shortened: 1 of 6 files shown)


db/migrate/20180204235116_add_mooc_to_code_schools.rb
```diff
@@ -0,0 +1,5 @@
+class AddMoocToCodeSchools < ActiveRecord::Migration[5.0]
+  def change
+    add_column :code_schools, :mooc, :boolean
```
hpjaj: Let's set a few things in this migration: 

1. default this column to `false`.  That way, all existing records will be set to false, and the frontend can immediately begin filtering data
2. set `null` to `false`.  This will enforce presence, at the database level.

Ping me if you have any questions on setting these!

leenyburger: @hpjaj If I validate the new attribute, won't that break our existing CodeSchool database?

hpjaj: @leenyburger - It would if we were not also setting a default of `false`.  But you're good, as long as that default is included, as well.

apiary.apib
```diff
@@ -50,6 +51,7 @@ API endpoints that Operation Code's Rails backend makes available to its React f
     + has_online (boolean, required) - Does the CodeSchool have an online component
     + online_only (boolean, required) - Is the CodeSchool online only?
     + notes (text, optional) - Any additional information about the CodeSchool
+    + mooc (boolean, required) - Does the CodeSchool offer Massive Open Online Courses?
```
hpjaj: These will actually be `optional`.  Since we are setting a default of `false` at the db level, this is already accounted for.

test/factories/code_schools.rb
```diff
@@ -8,5 +8,6 @@
     has_online Faker::Boolean.boolean
     online_only Faker::Boolean.boolean
     notes Faker::Lorem.paragraph
+    mooc Faker::Boolean.boolean
```
hpjaj: To make this factory a bit more performant, we can just set this to:

```ruby
mooc false
```

leenyburger:
@hpjaj There is a conflict in db/schema.rb. I think this means there was a migration added and merged after add_mooc_to_code_schools. I can't rebase on the command line because I'm getting an error: 
```
error: Failed to merge in the changes.
Patch failed at 0001 Added boolean mooc to code schools table
The copy of the patch that failed is found in: .git/rebase-apply/patch
```

hpjaj:
@leenyburger - Your rebase issues should be resolved now.  You can pull down the fresh changes from GitHub.

You'll want to check (`make test`) and fix all the failing specs, based on the new `mooc` column being required now.

Ping me if you have any issues!

kylemh:
I don't see `mooc` in the [API response for code_schools](api.operationcode.org/api/v1/code_schools)

hpjaj:
@kylemh - our bad.  resolved in https://github.com/OperationCode/operationcode_backend/pull/269

Issue #255, opened 2018-02-26
Code Platoon VA Approved!


JYoung217:
# Feature
## Why is this feature being added?
Code Platoon was approved by the VA on Friday for the GI Bill. We are updating our records with Operation Code to reflect that.

## What should your feature do?
We should show up under the "VA-Approved Schools" section and all the cards where Code Platoon is featured in should have "GI Bill Accepted: True" written.

hpjaj:
@JYoung217 - I updated this in prod's db!

PR #256, opened 2018-02-26
remove incorrect note section, add that we are a VA approved code school

JYoung217:
# Description of changes
This solves https://github.com/OperationCode/operationcode_backend/issues/255. Code Platoon was approved for the GI Bill and we want to reflect that in our entry with Operation Code.


hpjaj:
I updated this in prod's db!

PR #285, opened 2018-03-27
Documentation

dave5801:
# Description of changes
I am addressing Issue#271, documentation for API endpoints. 

I have changed the file apiary.ipib to account for:

api/v1/mentors_controller#index
api/v1/mentors_controller#show
api/v1/services_controller#index
api/v1/status_controller#all

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -1043,6 +1043,16 @@ API endpoints that Operation Code's Rails backend makes available to its React f
             errors: "Some error message"
         }
 
+## Status | Collection [/api/v1/status]
+
+### Get all Status Codes [GET]
+
++ Response 200 (application/json)
+
+    {
+        "status": ok 
+    }
+
 ## Tag | Collection [/api/v1/tags]
 
 ### List All Tags [GET]
--- a/apiary.apib
+++ b/apiary.apib
@@ -1049,9 +1049,17 @@ API endpoints that Operation Code's Rails backend makes available to its React f
 
 + Response 200 (application/json)
 
-    {
-        "status": ok 
-    }
+    + Body
+
+            {
+                status: "ok"
+            }
+
++ Response 422 (application/json)
+
+        {
+            errors: "Some error message"
+        }
 
 ## Tag | Collection [/api/v1/tags]
 
--- a/apiary.apib
+++ b/apiary.apib
@@ -734,6 +734,12 @@ API endpoints that Operation Code's Rails backend makes available to its React f
         {
             errors: "Some error message"
         }
+        
+## Mentor | Show [/api/v1/code_schools/{mentor_user_id}]
++ Parameters
+
+    + mentor_user_id (number) - ID of the Mentor in the form of an integer
+
 ## Resource [/api/v1/resources/{resource_id}]
 
 + Parameters
--- a/apiary.apib
+++ b/apiary.apib
@@ -1018,6 +1018,23 @@ API endpoints that Operation Code's Rails backend makes available to its React f
                 errors: "Some error message"
             }
 
+## Services | Collection [/api/v1/services]
+
+### List All Services [GET]
+
++ Response 200 (application/json)
+
+    + Body
+
+            {
+                service: "name"
+            }
+
++ Response 422 (application/json)
+
+        {
+            errors: "Some error message"
+        }
 
 ## Slack User | Invite [/api/v1/slack_users]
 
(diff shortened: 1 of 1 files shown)


apiary.apib
```diff
@@ -719,12 +719,35 @@ API endpoints that Operation Code's Rails backend makes available to its React f
 
 + Response 400 (application/json)
 
+### List All Mentors [GET]
```
hpjaj: These `Mentor` endpoints will go under their own h2 heading, and alphabetized accordingly.

So not under `Requests`, but under `Mentors`.

apiary.apib
```diff
@@ -719,12 +719,35 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
+
++ Response 200 (application/json)
+
+    + Body
+
+            {
+                "mentor": name
```
hpjaj: To get a sense of what is returned from a given endpoint response, you can:

1. `byebug` into the associated test
2. check what the `response.body` looks like

For example, for the index action you could:

1.  in here  https://github.com/OperationCode/operationcode_backend/blob/master/test/controllers/api/v1/mentors_controller_test.rb#L5-L11
2. comment out the `skip`
3. add a `byebug` below line 10
4. call `response.body`


And then you would follow the same general pattern for all the other endpoints you are adding documentation for.

hpjaj: @dave5801 - This response is what needs updating.  Right now it is showing up as:

```
{
    "mentor": name
}
```

Same applies to the "services collection" and the "status collection" endpoints.

hpjaj:
Also:

image

apiary.apib
```diff
@@ -541,6 +541,80 @@ API endpoints that Operation Code's Rails backend makes available to its React f
                 status: "ok"
             }
 
++ Response 422 (application/json)
+
+        {
+            errors: "Some error message"
+        }
+
+## Mentor | Collection [/api/v1/mentors/{?id,email,zip,first_name,last_name,slack_name,mentor,verified}]
```
hpjaj: When you are passing parameters to a `GET`, you will not have a `/` that separates them from the main URL.   So you'll remove the `/` between `mentors` and `{`.

apiary.apib
```diff
@@ -541,6 +541,80 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
+    + mentor (boolean, required) - Has the mentor been verified in the form of a boolean
+    + verified (boolean, required) - Has the mentor not been verified in the form of a boolean
+
+### List All Mentors [GET]
+
++ Response 200 (application/json)
+
+    + Body
+
+            {
```
hpjaj: This endpoint returns a collection of `Mentor` records.  So this will be an **array** of hashes.

apiary.apib
```diff
@@ -541,6 +541,80 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
+            },
+
+
++ Response 422 (application/json)
+
+        {
+            errors: "Some error message"
+        }
+ 
+## Mentor | Show [/api/v1/requests/{?id,email,zip,first_name,last_name,slack_name,mentor,verified,bio}]
```
hpjaj: The only parameter a `show` action will take is the `id` of the record to be shown.  

Here is an example for reference:  https://operationcodeapi.docs.apiary.io/#reference/0/resource-show-or-delete/view-a-resource's-detail

apiary.apib
```diff
@@ -1019,6 +1091,24 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
 
+## Services | Collection [/api/v1/services]
+
+### List All Services [GET]
+
++ Response 200 (application/json)
+
+    + Body
+
+            {
```
hpjaj: Same thing here re: an array of hashes

apiary.apib
```diff
@@ -1043,6 +1133,25 @@ API endpoints that Operation Code's Rails backend makes available to its React f
...
+
+## Status | Collection [/api/v1/status]
+
+### Get all Status Codes [GET]
+
++ Response 200 (application/json)
+
+    + Body
+
+            {
```
hpjaj: Same thing here re: an array of hashes

hpjaj:
@dave5801 - How goes?  Can we get an update on your progress with your PR?  Are you blocked on anything?

dave5801:
@hpjaj I was busy preparing for interviews -- I'll get back to this project when I can.

PR #292, opened 2018-04-07
Create alert email for leaders if someone signs up within 50 miles

### LINKED ISSUE #37: Alert community leaders when a new member joins in their area
# Feature

Community leaders have already become a vital part of Operation Code. They provide local support, connections and a friendly face near to our users. We should make it easier for our community leaders to know when a new member joins in their area.

When a member joins check to see if there's a community leader within a certain range of the new user. If there is send the community leader a message notifying of the new user.

nagano564:
# Description of changes

This should send an email to leaders if a new user signs up within 50 miles of the leader

# Issue Resolved

Fixes #37


app/models/user.rb
```diff
@@ -27,6 +29,7 @@ class User < ApplicationRecord
   before_validation :geocode, if: ->(v) { v.zip.present? && v.zip_changed? }
   before_save :upcase_state
   before_save :downcase_email
+  after_commit :notify_leaders_on_geocode_update, on: [:create, :update]
```
hpjaj: Since this is exclusive to both `:create` and `:update`, I believe you can just truncate this to just `on: :save`

app/views/user_mailer/new_user_in_leader_area.html.erb
```diff
@@ -0,0 +1 @@
+<%= @new_user.name %>, has joined Please email them at <%= @user_user.email %> and welcome them.
```
hpjaj: Please connect and collaborate with @hollomancer, @kylemh and @jenn (not sure of Jenn's GitHub handle) on this mailers:

- subject
- body

test/jobs/send_email_to_leaders_job_test.rb
```diff
@@ -0,0 +1,13 @@
+require 'test_helper'
+
+class SendEmailToLeadersJobTest < ActiveJob::TestCase
+  test 'send email to leaders if someone signs up within X miles' do
+
+    user = build(:user, email: 'static@email.com', latitude: 40.7, longitude: -73.8)
+    leader_user = create(:user, latitude: 40.7001, longitude: -73.8001)
+    leader_user.tag_list.add('community-leader')
+    leader_user.save
+    UserMailer.expects(:new_user_in_leader_area).with(leader_user, user)
```
hpjaj: Ensure that there is a passed `user_id`, and the associated `User` is found

ramillim:
@nagano564 I rewrote the test because I discovered that running a .count on the ActiveRecord relation after the geo-query causes the query to break (appeared to be a table alias related to the Geocoder gem).

Before we reconvene, make sure you git pull the last commit I pushed and then 'git fetch upstream' and rebase it with master. @hpjaj added a constant for the 'community leader' tag and a has_tag? method on User model.

app/jobs/send_email_to_leaders_job.rb
```diff
@@ -0,0 +1,11 @@
+class SendEmailToLeadersJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(user)
+    User.near([user.latitude, user.longitude],50)
```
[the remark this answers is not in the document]
ramillim: We renamed this to community_leaders_nearby(latitude, longitude, radius) since we had to change the method signature to pass in latitude, longitude.

There was an alternate way to do this I noticed after re-reading the doc (https://github.com/alexreisner/geocoder). We could have made this an instance method and leveraged Geocoder's .nearbys to avoid passing the lat and long, so it could be something like

```
def community_leaders_within(radius)
  nearbys(radius).tagged_with(User::LEADER)
end
```
I kind of like the latter using nearbys. The class method lets you pass in arbitrary coordinates, but we don't have a use case for that yet.

app/models/user.rb
```diff
@@ -84,6 +86,12 @@ def self.count_by_location(location, radius=20)
     near(location, radius.to_i)&.size
   end
 
+  def self.community_leaders_nearby(latitude, longitude, radius)
+    self
+      .near([latitude, longitude], radius)
+      .tagged_with('community-leader')
```
hpjaj: You can pass in the the constant here for `LEADER`:

https://github.com/OperationCode/operationcode_backend/blob/master/app/models/user.rb#L2

app/views/user_mailer/new_user_in_leader_area.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= @new_user.name %>, has joined. They're located within 50 miles of your chapter! Please email them at
+<%= @user_user.email %> and welcome them, and consider providing info on the next local meetup.
+Let's encourage in-person involvement within Operation Code!
```
hpjaj: Nice.  Did you confirm this email's subject and body text with @hollomancer @kylemh and/or @jenn?

test/models/user_test.rb
```diff
@@ -169,13 +191,31 @@ def user_opts
     refute "john@gmail" =~ User::VALID_EMAIL
     refute "@example.com" =~ User::VALID_EMAIL
   end
+
+  test '.community_leaders_nearby returns leaders within a radius from a lat/long' do
+    User.any_instance.stubs(:geocode)
+    tom = create :user, latitude: 1, longitude: 1
+    tom.tag_list.add('community-leader')
```
hpjaj: You can use the constant here, too.  To ensure data integrity.

hpjaj:
Also looks like CI turned up a couple errors.  You can click the Travis CI link above, and search for "errors" to see the issues.

You can also run `make test` locally, to see the errors.

ramillim:
@hpjaj We were just checkpointing it earlier today since we had to work on other things. Currently pairing on slack's screenhero right now, adding more tests for UserMailer right now.

app/views/user_mailer/new_user_in_leader_area.html.erb
nagano564: Yes me and @kylemh talked about the body last night. I created the subject.

nagano564:
Should we send each leader a email individually

Or have one email sent to all leaders in the area
    if we send it this way should we worry about having all the emails on the BCC line so other leaders can't see each others email addresses in the to line.

hpjaj:
This looks good!

> Should we send each leader a email individually

Yes.

We need a `rake` task to initially tag the appropriate users in prod with the community leader tag.  This would involve:

1. Collaborating with @hollomancer (or whomever he deems appropriate) to determine what prod users should be tagged
2. Creating the `rake` task

Would you like to include that as part of this PR and issue?  Or create a new issue and PR for it?

hpjaj:
Creating a separate issue for the `rake` task.

PR #316, opened 2018-04-20
Add is_partner to the CodeSchool table

### LINKED ISSUE #275: Add `is_partner` boolean attribute to CodeSchool table
# Feature

The frontend uses this descriptor for filtering on the code schools page.   We rarely seem to remove partner schools.  As such, it would make filtering easier if this attr was available as part of the endpoints responses, by default.

- [x] Add `is_partner` to the `CodeSchool` table.  Defaulting to `false` and `null` `false`
- [x] Add attr to serializer
- [x] Add attr to strong params
- [x] Update apiary docs
- [ ] Add spec illustrating presence
- [x] Add attr to factory, default to `false`
- [ ] Creates a new `rake` task to set any existing, relevant `CodeSchool` records in prod to `is_partner: true`

nagano564:
# Description of changes
 Add is_partner to the CodeSchool table. Defaulting to false and null false
 Add attr to serializer
 Add attr to strong params
 Update apiary docs
 Add spec illustrating presence
 Add attr to factory, default to false
 Creates a new rake task to set any existing, relevant CodeSchool records in prod to is_partner: true

# Issue Resolved

Fixes #275

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -70,12 +70,12 @@ API endpoints that Operation Code's Rails backend makes available to its React f
                       "name": "CoderSchool",
                       "url": "http://www.CoderSchool.vn",
                       "logo": "https://d388w23p6r1vqc.cloudfront.net/img/startups/11852/logo-1457327647.png",
-                      "full_time": "false",
-                      "hardware_included": "false",
-                      "has_online": "true",
-                      "online_only": "false",
+                      "full_time": false,
+                      "hardware_included": false,
+                      "has_online": true,
+                      "online_only": false,
                       "is_partner": false,
-                      "mooc": "false"
+                      "mooc": false
                       }
                   }
 
--- a/apiary.apib
+++ b/apiary.apib
@@ -20,6 +20,7 @@ API endpoints that Operation Code's Rails backend makes available to its React f
                 "hardware_included": false,
                 "has_online": false,
                 "online_only": false,
+                "is_partner": false,
                 "mooc": false
                 "locations":  [
                       {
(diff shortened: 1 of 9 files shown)


apiary.apib
```diff
@@ -71,7 +73,8 @@ API endpoints that Operation Code's Rails backend makes available to its React f
                       "full_time": "false",
                       "hardware_included": "false",
                       "has_online": "true",
-                      "online_only": "false"
+                      "online_only": "false",
+                      "is_partner": false,
```
hpjaj: Can you pls change all of these `true/false` in this endpoint from a string to a boolean (like yours).

lib/tasks/update_is_partner_for_code_schools.rake
```diff
@@ -0,0 +1,8 @@
+namespace :is_partner do
+  desc "Updates column is_partner for code schools"
+  task code_schools: :environment do
```
hpjaj: The `rake` task would be similar to the one you did for community leaders.  In the sense that you'll need to figure out which code schools, in prod, need to have this set to `is_partner: true`.

And then we would update just those specific schools.

To get started on that list, perhaps connect with @ksmacleod99 .  She might be able to point you in the right direction.

test/models/code_schools_test.rb
```diff
@@ -0,0 +1,12 @@
+require 'test_helper'
+
+class CodeSchoolTest < ActiveSupport::TestCase
+  def setup
+    @code_school = build :code_school
+  end
+
+  test "is_partner should be present" do
+    @code_school.is_partner = true
+    assert @code_school.valid?
```
hpjaj: As you updated the factory to set `is_partner` by default, you should be able to:

1. Create a new code school
2. Verify that `is_partner` is set to `false` (i.e. `assert @code_school.is_partner == false`)

hpjaj:
@nagano564 - Here is a current list of the code schools in prod:

https://gist.github.com/hpjaj/517f132bed58f3226b4728744578a823

nagano564:
Here is a gist of the current school that is_partner

https://gist.github.com/nagano564/b0af55dae73e6e4249d29698430c2c0f

PR #324, opened 2018-04-22
Slack Integration 1: Users and SlackUser db migration

wimo7083:
# Description of changes

1.  Add's `slack_users` table with index for both `slack_id` and `slack_email`.
 a. slack_id doesn't change and is a good candidate for using with the update route, since I'm not sure if they send the old and new values. 
b. slack_email is a good candidate to index when performing the create action. 

2. Dropped `slack_name` from the `users` table. 
3. Updated the `user` serializer, will probably update to include join when data is populated. 

https://gist.github.com/wimo7083/cd8fe53fc520621cf713da3ebb23a91a#file-make-db_migrate


db/migrate/20180422030623_create_slack_user.rb
```diff
@@ -0,0 +1,11 @@
+class CreateSlackUser < ActiveRecord::Migration[5.0]
+  def change
+    create_table :slack_users do |t|
+      t.string :slack_id, index: true
+      t.string :slack_name
+      t.string :slack_real_name
+      t.string :slack_display_name
+      t.string :slack_email, index: true, foreign_key: true
```
hpjaj: We won't have a foreign key here.  Our association will be through `user`.

db/migrate/20180422030623_create_slack_user.rb
```diff
@@ -0,0 +1,11 @@
+class CreateSlackUser < ActiveRecord::Migration[5.0]
+  def change
+    create_table :slack_users do |t|
+      t.string :slack_id, index: true
```
hpjaj: You'll need to add a column for the `user` association.  i.e.

```ruby
t.integer :user_id
```

db/migrate/20180422030623_create_slack_user.rb
```diff
@@ -0,0 +1,11 @@
...
+      t.string :slack_name
+      t.string :slack_real_name
+      t.string :slack_display_name
+      t.string :slack_email, index: true, foreign_key: true
+    end
```
hpjaj: You'll need to add in the timestamp columns.  Here is an example of that:

https://github.com/OperationCode/operationcode_backend/blob/master/db/migrate/20170920093119_create_code_schools.rb#L12

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180401034441) do
+ActiveRecord::Schema.define(version: 20180422031009) do
```
hpjaj: Heads up, before you make any changes to the `SlackUser` migration file, you will need to roll back these two migrations.

The way to do that with docker is [make db_rollback](https://github.com/OperationCode/operationcode_backend/blob/master/Makefile#L46-L48). Running this command will rollback the *most recent* migration.  So you will need to run it twice, in order to get to the `SlackUser` migration.

Then you'll make your changes in the migration files.  And then run `make db_migrate` to run them, and update the `schema`.

db/migrate/20180422030623_create_slack_user.rb
wimo7083: Added this column. I included index. Is that what you meant?

wimo7083:
Added the presence of validations for the items that can come from the pybot, didn't include `created_at` or `updated_at`.

Not sure if this is what you meant but these should be the only values that would be coming in.

db/schema.rb
wimo7083: I did the rollback 2x, changed the schema, then performed `db_migrate`.

hpjaj:
At first glance this looks good.  After you've fixed the merge conflicts I'll pull it down and take another look.

wimo7083:
Thanks for the assist.

PR #333, opened 2018-04-27
Swap existing Pundit implementation for CanCanCan

### LINKED ISSUE #301: Swap existing Pundit implementation for CanCanCan
# Feature

Pundit is only used in one place in our app.  CanCanCan works with ActiveAdmin, and supports a tiered permissions approach.

- [ ] Find where Pundit is used and swap it out for CanCanCan
- [ ] Remove Pundit from the app

nagano564:
# Description of changes
  Find where Pundit is used and swap it out for CanCanCan
  Remove Pundit from the app

# Issue Resolved
Fixes #301


nagano564:
I was also thinking of deleting test/factories/requests.rb but wanted to get confirmation before I did it.

nagano564:
so I am getting this error message

Error:
Api::V1::RequestsControllerTest#test_show_individual_request:
NoMethodError: undefined method `authorize!' for #<Api::V1::RequestsController:0x00000003fd77c8>
    app/controllers/api/v1/requests_controller.rb:18:in `show'
    test/controllers/api/v1/requests_controller_test.rb:40:in `block in <class:RequestsControllerTest>'
    test/test_helper.rb:39:in `instance_eval'
    test/test_helper.rb:39:in `block (2 levels) in test'
    test/test_helper.rb:38:in `block in test'
 
which I don't understand because the `authorize!` method should be build in 

@hpjaj

nagano564:
to test it I inserted

load_and_authorize_resource

to the request class

Error:
Api::V1::RequestsControllerTest#test_mentors_can_see_all_unclaimed_requests:
ActionController::RoutingError: undefined local variable or method `load_and_authorize_resource' for Api::V1::RequestsController:Class
    app/controllers/api/v1/requests_controller.rb:5:in `<class:RequestsController>'
    app/controllers/api/v1/requests_controller.rb:3:in `<module:V1>'
    app/controllers/api/v1/requests_controller.rb:2:in `<module:Api>'
    app/controllers/api/v1/requests_controller.rb:1:in `<top (required)>'
    test/controllers/api/v1/requests_controller_test.rb:11:in `block in <class:RequestsControllerTest>'
    test/test_helper.rb:39:in `instance_eval'
    test/test_helper.rb:39:in `block (2 levels) in test'
    test/test_helper.rb:38:in `block in test'

and this was the error. I also tried to byebug into the ability class and it was ever invoked which I think means that it doesn't load. but when trying load 'cancancan' with the byebug in the method it states false

app/controllers/api/v1/requests_controller.rb
```diff
@@ -4,7 +4,7 @@ class RequestsController < ApiController
       before_action :authenticate_user!
 
       def index
-        authorize Request
+        authorize! Request
```
hpjaj: Try `authorize! :read, Request`

app/controllers/api/v1/requests_controller.rb
```diff
@@ -15,7 +15,7 @@ def create
 
       def show
         @request = Request.find_by(id: params[:id])
-        authorize @request
+        authorize! @request
```
hpjaj: Try `authorize! :read, @request`

app/models/ability.rb
```diff
@@ -23,6 +25,8 @@ def initialize(user)
 
       return unless user.role&.super_admin?
       can :manage, :all
+    elsif user.mentor?
```
hpjaj: We'll want all persisted users to drop into this conditional, instead of qualifying it with an `elsif`.  So for example, 

```ruby
else
  return unless user.mentor?
  can :read, Request
```

Once we implement roles, we will take another look at this.  But for now, we are only concerned with mentors.

app/controllers/api/v1/requests_controller.rb
ramillim: After inspecting the ancestors of ApiController, the CanCan mixin module is not included. It only shows up in ApplicationController https://github.com/CanCanCommunity/cancancan/issues/217

ramillim: Trying to figure out which configuration needs to be set so that ApiController also gets CanCan::ControllerAdditions

ramillim:
I'm thinking that you're just going to have to modify ApiController
```ruby
class ApiController < ActionController::API
  include CanCan::ControllerAdditions

  wrap_parameters format: [:json]
end
```
The reason why is because the way the CanCan:ControllerAdditions mixin gets included is with this
https://github.com/CanCanCommunity/cancancan/blob/master/lib/cancan/controller_additions.rb#L387
```ruby
ActiveSupport.on_load(:action_controller) do
  include CanCan::ControllerAdditions
end
```

And since ApiController inherits from ActionController::API instead of ActionController, it doesn't get added :(

Or you could contribute to the cancancan gem to be able to add the option :)

app/controllers/api/v1/requests_controller.rb
hpjaj: Good call @ramillim 👍 

Let's go with adding `include CanCan::ControllerAdditions` to `app/controllers/api_controller.rb`.

ramillim:
@hpjaj @nagano564 
What do you think about handling the exception like this? (following convention from other API Controllers such as ScholarshipController). I verified that when the Pundit gem was being used, it was an unhandled exception.

BTW, the default error message for this exception is "CanCan::AccessDenied: You are not authorized to access this page." so a little bit weird that it says page, so I set a new default below. You can override the message on a per-action basis in the authorize! call.

```
class ApiController < ActionController::API
  include CanCan::ControllerAdditions
  wrap_parameters format: [:json]

  rescue_from CanCan::AccessDenied do |exception|
    exception.default_message = 'You are not authorized to access this resource.'
    render json: { error: exception.message }, status: :forbidden
  end
end
```

test/controllers/api/v1/requests_controller_test.rb
```diff
@@ -13,6 +12,14 @@ class Api::V1::RequestsControllerTest < ActionDispatch::IntegrationTest
     assert_equal 1, response.parsed_body.count
   end
 
+  test "users that are not mentors can't see all unclaimed requests" do
```
hpjaj: Can you pls add a similar test for the `:show` action

apiary.apib
```diff
@@ -673,6 +673,12 @@ API endpoints that Operation Code's Rails backend makes available to its React f
             error: 'Unauthorized'
         }
 
++ Response 403 (application/json)
```
hpjaj: Perfect.  Pls add for the `:show` action, as well.

app/models/ability.rb
hpjaj: This, too

PR #334, opened 2018-04-29
adds in instructions for entering rails console with local docker setup

nellshamrell:
# Description of changes
When running the local dev setup, I realized I wasn't sure how to enter the Rails console.  I figured it out and thought I should add it to the docs :)

### DIFF
--- a/docs/setup/docker_setup.md
+++ b/docs/setup/docker_setup.md
@@ -82,5 +82,3 @@ $ docker exec -it [CONTAINER-ID] bash
 root@[CONTAINER-ID]:/app# rails console
 ```
 * And you should be in the Rails console!
-
-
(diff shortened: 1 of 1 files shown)


kirbygit:
docker-compose run web rails console <-- is this the right command?

hpjaj:
@nellshamrell - For local dev, to access the rails console in docker you can run `make console`.  Is that what you were talking about?

FYI, there are a handful of commands like this, that all can be found in the `Makefile`.  Here is the one for the rails console:

https://github.com/OperationCode/operationcode_backend/blob/master/Makefile#L26-L28

nellshamrell:
Ah!  that is much better!  Closing this :)

PR #340, opened 2018-05-02
mark code platoon as va accepted

JYoung217:
# Description of changes
We're not listed as a school that accepts the GI bill, but we are. Simple `true` flag to flip this.


hpjaj:
Done

PR #355, opened 2018-05-13
Adds new resources to ActiveAdmin for `Scholarship` and `ScholarshipApplication`

### LINKED ISSUE #349: Adds new resources to ActiveAdmin for `Scholarship` and `ScholarshipApplication`
# Feature

Gives admins the ability to conduct CRUD actions on the `Scholarship` and `ScholarshipApplication` models.  Currently this has to be done through prod's rails console.

Here are the ActiveAdmin docs:  https://activeadmin.info/documentation.html

Here is a sample implementation of a Resource in our codebase:  https://github.com/OperationCode/operationcode_backend/blob/master/app/admin/team_member.rb

- [ ] Adds a Resource for `Scholarship` to ActiveAdmin
- [ ] Adds a Resource for `ScholarshipApplication` to ActiveAdmin
- [ ] Updates `ability.rb` with the proper permissions - confirm with @hpjaj 
- [ ] These two models have a `has_many/belongs_to` relationship with one another.  Be sure to take that into account.  We have an existing implementation for `CodeSchool` and `Location` that can be used for reference.
- [ ] Create three `AdminUser`s. One with each role: super admin, admin, and board member. And then login and out as each, testing access. This is all done in your local test env (i.e. localhost). The `seeds` file can help with this.

nagano564:
# Description of changes

  Gives admins the ability to conduct CRUD actions on the Scholarship and ScholarshipApplication 
  models. Currently this has to be done through prod's rails console.

  Adds a Resource for Scholarship to ActiveAdmin
  Adds a Resource for ScholarshipApplication to ActiveAdmin
  Updates ability.rb with the proper permissions 

# Issue Resolved

Fixes #349

### DIFF
--- a/app/admin/scholarship_application.rb
+++ b/app/admin/scholarship_application.rb
@@ -8,8 +8,8 @@ ActiveAdmin.register ScholarshipApplication do
       scholarship_app.scholarship.name
     end
 
-    column 'User' do |user|
-      user.scholarship.name
+    column 'User' do |user_id|
+      user_id.scholarship.name
     end
 
     column :reason
--- a/app/admin/scholarship_application.rb
+++ b/app/admin/scholarship_application.rb
@@ -23,6 +23,7 @@ ActiveAdmin.register ScholarshipApplication do
   form do |f|
     f.inputs do
       f.input :scholarship_id, label: 'Scholarship', as: :select, collection: Scholarship.all.order(:name).map { |scholarship| [scholarship.name, scholarship.id]}, include_blank: false
+      f.input :user_id, label: 'User', as: :select, collection: User.all.order(:name).map { |user| [user.name, user.id]}, include_blank: false
       f.input :reason
       f.input :terms_accepted
     end
--- a/app/admin/scholarship_application.rb
+++ b/app/admin/scholarship_application.rb
@@ -22,13 +22,12 @@ ActiveAdmin.register ScholarshipApplication do
 
   form do |f|
     f.inputs do
-      f.input :scholarship_id, label: 'Scholarship', as: :select, collection: Scholarship.all.order(:name).map { |scholarship| [scholarship.name, scholarship.id]}, include_blank: false
-      f.input :user_id, label: 'User', as: :select, collection: User.all.order(:name).map { |user| [user.name, user.id]}, include_blank: false
+      f.input :scholarship_id, label: 'Scholarship', as: :select, collection: Scholarship.all.order(:name).map {|scholarship| [scholarship.name, scholarship.id]}, include_blank: false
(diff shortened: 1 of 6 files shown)


app/admin/scholarship_application.rb
```diff
@@ -0,0 +1,35 @@
...
+  index do
+    selectable_column
+    id_column
+
+    column 'Scholarship' do |scholarship_app|
+      scholarship_app.scholarship.name
+    end
+
+    column 'User' do |user_id|
+      user_id.user.email
```
hpjaj: Can you please create a method in `app/models/user.rb` for a user's full name that returns of string of their first and last name, and then use that method here, in place of `email`.

app/admin/scholarship_application.rb
```diff
@@ -0,0 +1,35 @@
...
+    column :created_at
+    column :updated_at
+
+    actions
+  end
+
+  form do |f|
+    f.inputs do
+      f.input :scholarship_id, label: 'Scholarship', as: :select, collection: Scholarship.all.order(:name).map { |scholarship| [scholarship.name, scholarship.id] }, include_blank: false
+      f.input :user_id, label: 'User', as: :select, collection: User.all.order(:email).map { |user| [user.email, user.id] }, include_blank: false
```
hpjaj: Here we can:

- `order` by `last_name`
- use the new `full_name` method in place of `email` in the array

That way the dropdown will be sorted by last name (there are thousands of users), and the dropdown itself will contain the user's full name.

test/factories/users.rb
```diff
@@ -12,7 +12,7 @@
     bio { Faker::Company.bs }
     timezone 'EST'
     mentor false
-    verified false
+    verified true
```
hpjaj: Let's keep this set to `false` by default, as this is not the initial sign on case until someone goes through the ID.me process.

If you have need it to be `true`, you can either:

- use the `trait`:  https://github.com/OperationCode/operationcode_backend/blob/master/test/factories/users.rb#L26-L28
- set it to `true` in the test in question

test/factories/scholarship_applications.rb
```diff
@@ -1,8 +1,7 @@
 FactoryGirl.define do
   factory :scholarship_application do
     reason 'MyText'
-    terms_accpeted false
-    references ''
-    references ''
+    terms_accepted false
+    user
```
hpjaj: We will need this user to be a `verified: true` user.  Can you check out the FactoryGirl docs for the proper way to insert that here?

test/factories/scholarship_applications.rb
```diff
@@ -1,8 +1,7 @@
 FactoryGirl.define do
   factory :scholarship_application do
     reason 'MyText'
```
hpjaj: This will also require an associated `Scholarship`, similar to how this needs an associated `User`.

PR #366, opened 2018-06-01
Issue365 add jobs

### LINKED ISSUE #365: Add jobs table, endpoint, and functionality
# Feature
Allow jobs to be saved in the database and provided to the front end via an endpoint. This feature will also allow CRUD actions via the admin dashboard. 

Needs to include: 
- new db table
- new `GET` index endpoint 
- new resource in ActiveAdmin
- test coverage
- API docs

Jobs table to include: 
title
source_url
source
city
state
country
description
status (i.e. active, inactive)
remote 
Also include tagging. 
---

# Bug Report

leenyburger:
# Description of changes
Add job model, controller, factory, specs, admin interface, apiary docs

# Issue Resolved

Fixes #365

### DIFF
--- a/app/admin/job.rb
+++ b/app/admin/job.rb
@@ -1,0 +1,23 @@
+ActiveAdmin.register Job do
+# See permitted parameters documentation:
+# https://github.com/activeadmin/activeadmin/blob/master/docs/2-resource-customization.md#setting-up-strong-parameters
+#
+permit_params :title, :source_url, :source, :city, :state, :country, :description, :status, :remote
+
+  index do
+    selectable_column
+    id_column
+
+    column :title
+    column :source_url
+    column :source
+    column :city
+    column :state
+    column :country
+    column :description
+    column :status
+    column :remote
+
+    actions
+  end
+end
--- a/app/controllers/api/v1/users/jobs_controller.rb
+++ b/app/controllers/api/v1/users/jobs_controller.rb
@@ -1,0 +1,11 @@
+module Api
+  module V1
+    class JobsController < ApiController
+      def index
+        render json: Job.all, status: :ok
+      rescue StandardError => e
+        render json: { errors: e.message }, status: :unprocessable_entity
+      end
+    end
+  end
+end
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -1,0 +1,2 @@
+class Job < ApplicationRecord
+end
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -1,2 +1,3 @@
 class Job < ApplicationRecord
+  acts_as_taggable_on
 end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -31,6 +31,9 @@ FactoryGirl.create(:request, assigned_mentor: nell)
   Service.create!(:name => service)
 end
 
+# Create jobs
+Job.create!(title: "A great job", source_url: "www.applyhere.com", source: "Company A", city: "Virginia Beach", state: "VA", country: "USA", description: "Our job is fun!", status: "active", remote: "false")
+
 # Create team members
 SeedTeamMembers.seed_all
 
--- a/test/factories/jobs.rb
+++ b/test/factories/jobs.rb
@@ -1,0 +1,13 @@
+FactoryGirl.define do
+  factory :job do
+    title Faker::Lorem.characters(7)
+    source_url Faker::Internet.url
+    source Faker::Lorem.characters(7)
+    city Faker::Address.city
+    state Faker::Address.state
+    country Faker::Address.country
+    description Faker::Lorem.paragraph
+    status "active"
+    remote false
+  end
+end
--- a/apiary.apib
+++ b/apiary.apib
@@ -435,6 +435,35 @@ API endpoints that Operation Code's Rails backend makes available to its React f
             errors: "Some error message"
         }
 
+## Jobs | Collection [/api/v1/jobs]
+
+## List all Jobs [GET]
+
++ Response 200 (application/json)
+
+        [
+          {
+            "id": 1,
+            "title": "A great job",
+            "source_url": "www.applyhere.com",
+            "source": "Company A",
+            "city": "Virginia Beach",
+            "state": "VA",
+            "country": "USA",
+            "description": "Our job is fun!",
+            "status": "active",
+            "remote": false,
+            "created_at": "2018-06-01T16:21:59.462Z",
+            "updated_at": "2018-06-01T16:21:59.462Z"
+          }
+        ]
+
++ Response 422 (application/json)
+
+        {
+          errors: "Some error message"
+        }
+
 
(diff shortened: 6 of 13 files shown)


jjhampton:
@leenyburger @hpjaj Any reason that `status` has to be a string? When using this data in the front-end, it seems a bit fragile to match on strings, especially if they're user-entered via Admin Dashboard, where there's a chance for human typos. Thoughts on using boolean instead?  Example: `active: true` or `active: false`.

app/controllers/api/v1/users/jobs_controller.rb
```diff
@@ -0,0 +1,11 @@
+module Api
```
hpjaj: Did this extra controller sneak in?

db/seeds.rb
```diff
@@ -31,6 +31,9 @@
   Service.create!(:name => service)
 end
```
hpjaj: Can you pls add a `delete_all` call for `Job` i.e. https://github.com/OperationCode/operationcode_backend/blob/master/db/seeds.rb#L9-L16

db/seeds.rb
```diff
@@ -31,6 +31,9 @@
...
 
+# Create jobs
+Job.create!(title: "A great job", source_url: "www.applyhere.com", source: "Company A", city: "Virginia Beach", state: "VA", country: "USA", description: "Our job is fun!", status: "active", remote: "false")
+
 # Create team members
```
hpjaj: As well as a reporting call, i.e.  https://github.com/OperationCode/operationcode_backend/blob/master/db/seeds.rb#L57-L64

jjhampton:
@hpjaj I've already coded up the front-end stuff to filter on for `status === "active"` for now, should be fine for the time being.
POC here is @ksmacleod99 - Please see above comments - do you envision ever needing to set a Job's status to anything other than just active / inactive?

app/models/job.rb
```diff
@@ -0,0 +1,3 @@
+class Job < ApplicationRecord
+  acts_as_taggable_on
```
hpjaj: Can you pls add a method similar to this one in here:

https://github.com/OperationCode/operationcode_backend/blob/master/app/models/user.rb#L130-L141

As well as an associated test.  This will confirm that a given job can have tags (plural) added to it, and then we can use this method to query for any jobs that match the search criteria.

app/admin/job.rb
```diff
@@ -0,0 +1,20 @@
...
+    column :source
+    column :city
+    column :state
+    column :country
+    column :description
+    column :status
+    column :remote
+
+    actions
+  end
```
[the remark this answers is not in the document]
leenyburger: When I add the form and try to add a job I get the following error: "  "exception": "#<ActionView::Template::Error: To use the :country input, please install a country_select plugin, like this one: https://github.com/stefanpenner/country_select>",
Should I install that gem, or are we already using something else?

hpjaj:
re: @jjhampton 's comment here:  https://github.com/OperationCode/operationcode_backend/pull/366#issuecomment-394116169

I agree, would be easier to go the `boolean` route.  If that is still on the table (as I know it would require a bit of a refactor on the FE), I would recommend that, as it is for sure the cleanest solution.

If not, then @leenyburger you'll want to follow this pattern in order to implement choices for `"active"` and `"inactive":

https://github.com/OperationCode/operationcode_backend/pull/375#discussion_r202518607

After this piece is wrapped up, this PR should be ready to ship.

app/admin/job.rb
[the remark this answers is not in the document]
dmarchante: @hpjaj I don't see where you responded here. If @leenyburger issue has been addressed, disregard just getting visibility.

dmarchante:
Can you review the conflicts and address the final comments from @hpjaj @leenyburger?

leenyburger:
@dmarchante Yes. I'm out of town until next week but can take a look then.

wimo7083:
Hi @leenyburger how is this going? Is there anything blocking you that I can assist with?

leenyburger:
@wimo7083 No blockers, just haven't been able to find the time. Making up for hurricane delay next week, can take a look the week of the 24th.

leenyburger:
@wimo7083 I'm not going to be able to get to this in a timely manner. Sorry to leave you hanging!

wimo7083:
Okay so I'm going to make the decision we want to use a boolean instead of a string. I think this would align better with what the frontend needs.

For this I'd recommend:
1. changing `status` to `is_active_status`
2. changing all the mocks to respect a boolean instead of string
3. changing the schema to respect the new type. 

If nobody objects I'd like to finish this up.

robbkidd:
I have a maybe philosophical question: what does it mean for a job to have an "active status"?

kylemh:
@robbkidd Agreed. I feel like that field is unnecessary. I'd prefer a self-cleaning database.

We could end up having tons of inactive jobs

wimo7083:
Since we don't know when the job will expire. it's probably best to make it adjustable. I can add something to the route so we eliminate innactive jobs. 

I went with `is_open` because apparently using `open` in Factory girl gave me a security warning.

robbkidd:
🤔 One option for flagging jobs as either open or closed is to record the date on which it closed. This is a bit like a technique named "soft delete." It would involve adding a datetime column `closed_at` (if you wanted to match the Rails timestamps style). A nil `closed_at` means the job is still open; a non-nil `closed_at` means the job is closed and you also know the date it closed. Knowing the closed date means old-and-closed jobs could be removed from the system.

```

File: Makefile [hunk 6]
```diff
@@ -1,100 +1,108 @@
...
-bundle:
-	docker-compose run ${RAILS_CONTAINER} bash -c 'cd /app && bundle'
 
-setup: build db_create db_migrate
+#### Cleanup
+.PHONY: nuke
+nuke: 
+	$(DOCKER) system prune -a --volumes
+
+.PHONY:
```
robbkidd: `.PHONY: minty-fresh` here

test/factories/jobs.rb
```diff
@@ -0,0 +1,13 @@
+FactoryGirl.define do
+  factory :job do
+    title Faker::Lorem.characters(7)
+    source_url Faker::Internet.url
+    source Faker::Lorem.characters(7)
+    city Faker::Address.city
+    state Faker::Address.state
+    country Faker::Address.country
+    description Faker::Lorem.paragraph
```
robbkidd: All the attributes being set above this comment should get assigned with block to evaluate the Faker data lazily each time the factory is used instead of at test load time. Otherwise all jobs created with this factory will have the same values.

e.g. `title { Faker::Lorem.characters(7) }`

app/models/job.rb
```diff
@@ -0,0 +1,10 @@
+class Job < ApplicationRecord
```
wimo7083: Does having this in the models section of the code base make it clear enough that this is separate from Sidekiq Jobs?

BRIMIL01: Are you sure you don't want JobJobs?

robbkidd: Now that you mention it, maybe `WorkJobber`.

Issue #374, opened 2018-07-11
Formatting issue in db/schema.rb:14:1 on `make test`


hollomancer:
## What is the current behavior?
When running `make test`,

```
Inspecting 252 files                                                                                                                                 
.....................................................................................................................................................
....................................................C..................................................                                              
                                                                                                                                                     
Offenses:                                                                                                                                            
                                                                                                                                                     
db/schema.rb:14:1: C: Layout/EmptyLinesAroundBlockBody: Extra empty line detected at block body beginning.                                           
                                                                                                                                                     
252 files inspected, 1 offense detected                                                                                                              
make: *** [test] Error 1
```

aaron-suarez:
I have confirmed that running `make setup` consistently adds and extra line to `db/schema.rb` and adds an untracked file called `test/cassettes/user_test/military_status_accepts_valid_states.yml`. I'm on master 8fbb035f91d324b9d15d97977fa95bd551bf9d07

PR #375, opened 2018-07-11
Add military_status to user (fixes #373)

### LINKED ISSUE #373: Identify military spouses in backend admin interface
Not sure if this is a feature request or a bug.

# Bug Report
Our current user signup flow asks for information on whether someone is a military spouse, but this does not appear to be visible in the backend admin interface.

Ability to search by military spouse.

hollomancer:
# Description of changes
Adds a column for `military_status`, which is currently not present on `users`. This will allow the information we request regarding a military spouse / active duty in signup flow to save to the DB.
# Issue Resolved
Part of fixing #373 (some frontend work needed)

_It's been a while for me. Please review._

### DIFF
--- a/app/admin/user.rb
+++ b/app/admin/user.rb
@@ -114,7 +114,6 @@ ActiveAdmin.register User do
       f.input :linkedin
       f.input :employment_status
       f.input :education
-      f.input :military_status
       f.input :company_role
       f.input :company_name
       f.input :education_level
--- a/app/admin/user.rb
+++ b/app/admin/user.rb
@@ -106,6 +106,7 @@ ActiveAdmin.register User do
       f.input :volunteer
       f.input :branch_of_service
       f.input :years_of_service
+      f.input :military_status, as: :select, collection: User::MILITARY_STATUSES.map { |status| [status, status].compact }.reject(&:empty?), include_blank: true
       f.input :pay_grade
       f.input :military_occupational_specialty
       f.input :github
--- a/app/admin/user.rb
+++ b/app/admin/user.rb
@@ -6,7 +6,7 @@ ActiveAdmin.register User do
     :last_name, :timezone, :bio, :verified, :state, :address_1, :address_2, :city,
     :username, :volunteer, :branch_of_service, :years_of_service, :pay_grade,
     :military_occupational_specialty, :github, :twitter, :linkedin, :employment_status,
-    :education, :company_role, :company_name, :education_level, :interests
+    :education, :military_status, :company_role, :company_name, :education_level, :interests
 
   scope :all
   scope :mentors
@@ -74,6 +74,7 @@ ActiveAdmin.register User do
     column :linkedin
     column :employment_status
     column :education
+    column :military_status
     column :company_role
     column :company_name
     column :education_level
@@ -112,6 +113,7 @@ ActiveAdmin.register User do
       f.input :linkedin
       f.input :employment_status
       f.input :education
+      f.input :military_status
       f.input :company_role
       f.input :company_name
       f.input :education_level
--- a/db/migrate/20180711212702_add_militarystatus_to_user.rb
+++ b/db/migrate/20180711212702_add_militarystatus_to_user.rb
@@ -1,0 +1,5 @@
+class AddMilitarystatusToUser < ActiveRecord::Migration[5.0]
+  def change
+    add_column :users, :military_status, :string
+  end
+end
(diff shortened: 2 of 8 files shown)


hollomancer:
Also fixes #374

db/migrate/20180711212702_add_militarystatus_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddMilitarystatusToUser < ActiveRecord::Migration[5.0]
+  def change
+    add_column :users, :military_status, :string
```
hpjaj: As this is a `string`, is the plan to accept any `string`, or a specific list of strings for consistency?  Meaning a specified set of four statuses, for example.  I **highly** recommend the latter.  

This way, we can enforce data integrity, hang logic on the responses, do data analysis, filtering and searching, add rules for certain statuses, etc., etc.

If the answer is yes, then we'll need to do a few extra additions in this PR before shipping it.

hollomancer: I'd just need to add the check in `/app/models/user.rb`, is that correct?

hpjaj: @hollomancer - To establish the specific set of validated statuses, you'll want to:

1. Create a constant for each status [here](https://github.com/OperationCode/operationcode_backend/blob/master/app/models/user.rb#L22).  
2. Create a constant to represent an array of all of the statuses.  You'll need to include one for `nil`, to account for our current users, and future users, that will not have a status.  For example:

```ruby
CURRENT = 'current'
VETERAN = 'veteran'
SPOUSE = 'spouse'
MILITARY_STATUSES = [CURRENT, VETERAN, SPOUSE, nil]
```

3. In `models/user.rb`, Add an `inclusion` validation for the new `:military_status` attribute, for the `MILITARY_STATUSES`.  [Here is how to do that.](http://guides.rubyonrails.org/active_record_validations.html#inclusion) 
4. In [`admin/user.rb`](https://github.com/OperationCode/operationcode_backend/blob/master/app/admin/user.rb#L93), add the set of selections to ActiveAdmin dashboard **form** portion.  For example:

```ruby
f.input :military_status, as: :select, collection: User::MILITARY_STATUSES.map { |status| [status, status].compact }.reject(&:empty?), include_blank: true
```

5. Add test coverage to `user_test.rb` testing the new inclusion validation
6. In [apiary.apib](https://github.com/hollomancer/operationcode_backend/blob/c0d852e8481d229d520652c7051567ff5af4ed7c/apiary.apib#L1810), in the `military_status` definition, change that to something like:

```
Users military status, either: current, veteran, spouse, or blank
```

dmarchante:
@hpjaj is this only waiting a review, or are there more changes to be made?

hollomancer:
@dmarchante, this one's on me to finish validation appropriately.

test/models/user_test.rb
```diff
@@ -1,6 +1,6 @@
 require 'test_helper'
 
-class UserTest < ActiveSupport::TestCase
+class UserTest < ActiveSupport::TestCase # rubocop:disable Metrics/ClassLength
```
hollomancer: @hpjaj, should we break this up into smaller segments?

hpjaj: No, let's keep all of the user model related test code in this file.  

As far as disabling RuboCop, you can:

- do it like this: http://www.mindonrails.com/notes/334-disabling-and-re-enabling-a-rubocop-rule-for-a-specific-set-of-code
- add an exclusion for the test file to `.rubocop.yml`

The latter is slightly preferable.

app/admin/user.rb
```diff
@@ -105,13 +106,15 @@
...
       f.input :years_of_service
+      f.input :military_status, as: :select, collection: User::MILITARY_STATUSES.map { |status| [status, status].compact }.reject(&:empty?), include_blank: true
       f.input :pay_grade
       f.input :military_occupational_specialty
       f.input :github
       f.input :twitter
       f.input :linkedin
       f.input :employment_status
       f.input :education
+      f.input :military_status
```
hpjaj: You'll want to delete this input (we'll favor the new collection input on line 109).

test/models/user_test.rb
```diff
@@ -260,4 +260,11 @@ def user_opts
     assert_equal data[:zip], userInfo[:zip]
     assert_equal '/profile', redirect
   end
+
+  test 'military status accepts valid states' do
+    assert create(:user, military_status: 'spouse')
```
hpjaj: Can you pls change these three to pass the three new constants, to support our validation?  i.e.

```ruby
assert create(:user, military_status: User::SPOUSE).valid?
```

Issue #384, opened 2018-08-23
Rubocop is too strict on windows


wimo7083:
# Bug Report
## What is the current behavior?
`make rubocop` should lint the files. Using Windows the carriage return characters for line endings make a lint offense. git normalizes line endings and shouldn't be considered. 

When I run lint I get 253 offenses detected.

## What is the expected behavior?
Use git line ending check to prevent, turn off line ending checks. 

### Operating System
Windows

### Browser
N/A
### Screenshots
![image](https://user-images.githubusercontent.com/10781353/44503963-56e44980-a656-11e8-8a36-23f0334b9812.png)

PR #380, opened 2018-08-13
Fix Mentor Request table tests

leenyburger:
# Description of changes
- Removed Airtable cassettes from old implementation 
- Recorded new VCR cassettes for the following files: 

1. `test/cassettes/airtable/client/get_records_successful.yml`
2. `test/cassettes/airtable/client/post_record_successful.yml`
3.  `test/cassettes/airtable/mentorship/post_multiple_successful.yml`
4. `test/cassettes/airtable/mentorship/post_successful.yml`

- Updated Apiary to include new email field
-  Updated Skillset options in test to account for new multiple choice options on Skillsets in Airtable

PR #391, opened 2018-08-30
Fixes few bugs

### LINKED ISSUE #390: Bug: Fix slack token
# Bug Report
We are failing to send slack invites, and this is being done manually. This is spamming monitoring with an except on every user signup. 

User gets invited. 

We need to get a new [admin legacy token](https://api.slack.com/custom-integrations/legacy-tokens) and install into environment variables. This token needs to only be used for this application.

wimo7083:
# Description of changes

Temporarily turns off leader emails due to bug with the sidekiq queue related to the database.
Temporarily turns off slack emails since @hollomancer loves to invite our new people (also a bug)
I also don't use git much, and I included a rubocop fix that should be separate (and it is), and my partner is wondering why I'm not spending time with her but I needed to get this fix out.

# Issue Resolved

Fixes #390 #380 #384

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,13 +5,17 @@ inherit_from: .rubocop_todo.yml
 # SupportedStyles: single_quotes, double_quotes
 Style/StringLiterals:
   Exclude:
-    - 'db/**/*'
+    - "db/**/*"
+
+#turn off end of line checks, git will fix this
+EndOfLine:
+  Enabled: false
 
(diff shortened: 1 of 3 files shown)


wimo7083:
Thanks for the help with this @nellshamrell 

To summarize:
1. remove the call to send email to leaders
2. remove the call to invite slack users
3. failing tests since we weren't doing the above two, so nell asked me to remove the tests
4. In the future we will remove connected code that is now dead.

PR #402, opened 2018-09-16
Fix send grid job

nellshamrell:
This fixes an error we were seeing in Sentry:

```
ActiveJob::DeserializationErrorSidekiq
errorError while trying to deserialize arguments: Couldn't find User with 'id'=5838
```

Turns out there is a common race condition when running a Sidekiq job right after a create or save - see [this extremely helpful article](https://www.justinweiss.com/articles/a-couple-callback-gotchas-and-a-rails-5-fix/) for more details.

I think that the current tests are sufficient to cover this behavior, but am willing to consider adding further testing if others think it is necessary.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -34,7 +34,7 @@ class User < ApplicationRecord
   #
   VALID_EMAIL = /\A([^@\s]+)@((?:[-a-z0-9]+\.)+[a-z]{2,})\z/i
 
-  after_save :welcome_user
+  after_commit :welcome_user
   before_validation :strip_zip_code
   before_validation :geocode, if: ->(v) { v.zip.present? && v.zip_changed? }
   before_save :upcase_state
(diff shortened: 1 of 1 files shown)


robbkidd:
@wimo7083 Yes to both.

We're almost there. Per [the Sidekiq FAQ about this error](https://github.com/mperham/sidekiq/wiki/FAQ#why-am-i-seeing-a-lot-of-cant-find-modelname-with-id12345-errors-with-sidekiq), we should also add a constraint to the registration of this callback to prevent welcoming users whenever their `User` record is modified.

```
after_commit :welcome_user, on: :create
```

### Possible Future ToDo:

A follow-on improvement to consider is to pass only `self.id` to `AddUserToSendGridJob.perform_later()` and update the job to do `user = User.find_by(id: user_id)`  first thing in the `perform()` method. This prevents serializing the entire `User` record into the job data to be stored in Redis. Sidekiq's documentation on [working with ActiveJob has words about this](https://github.com/mperham/sidekiq/wiki/Active-Job#limitations), but I've found life has been easier for debugging jobs by having the hand-off of data to a job to be as small as possible. Treat the `perform_later()` "scheduler" and the `perform()` "doer" as an API through which you pass only what's needed. This pattern makes more work for a job's `perform()` method and that's what gives you opportunities for error handling.

wimo7083:
Closing this due to continued work on : https://github.com/OperationCode/operationcode_backend/pull/407

PR #403, opened 2018-09-16
modify makefile to fix issue with console-sandbox target

wimo7083:
# Correct the makefile so the console-sandbox target works. 

Currently:
```
C:\Users\wimo7\Desktop\back [make_console_fix +1 ~0 -0 !]> make console-sandbox
make: Nothing to be done for 'console-sandbox'.
```

There is a target for this though:
https://github.com/OperationCode/operationcode_backend/blob/b34ad45cc77a0c86d3918681e83066cf5d58a060/Makefile#L31

While the `.PHONY` is properly declared there is no valid target due to a tab:
image

# Issue Resolved

Fixes using a makefile target to run the console with the operation code docker instance.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ fresh-restart: minty-fresh setup test run
 
 .PHONY: console-sandbox
 console-sandbox:
-	 docker-compose run ${RAILS_CONTAINER} rails console --sandbox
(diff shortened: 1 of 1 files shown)


wimo7083:
merging this while robb is waiting on the approval permission.

PR #409, opened 2018-09-25
remove validation of veteran status due to changed requirements

### LINKED ISSUE #408: Bug: Stop verifying ID.Me for scholarship applications.
# Bug Report
BE returns an error when an applicant wants to fill out a scholarship but they aren't verified.

Per leadership and issues with ID.me we don't want to use this anymore.

wimo7083:
# Description of changes

Removes the verification for users of id.me status. 

Removed the user_is_verified hook.
Removed the user_is_verified method
Removed the verified [delegate](https://apidock.com/rails/Module/delegate) - which is something I just learned about, but I still don't understand.

I left the FactoryGirl specifying a verified status, since it doesn't matter now. 

I haven't tested any verified routes before. This is a non-tested change, please don't merge unless you are able to test or assist me with testing this. 

I think I need to run both FE and BE locally, register a user and then test that this works locally. I'm going to try and get to that tomorrow. 

# Issue Resolved

Fixes #408

### DIFF
--- a/app/models/scholarship_application.rb
+++ b/app/models/scholarship_application.rb
@@ -1,5 +1,4 @@
 class ScholarshipApplication < ApplicationRecord
-  validate :user_is_verified
   validates :user_id, uniqueness: {scope: :scholarship_id}
 
   belongs_to :user
@@ -7,11 +6,4 @@ class ScholarshipApplication < ApplicationRecord
 
   delegate :name, to: :user, prefix: true
   delegate :email, to: :user, prefix: true
-  delegate :verified, to: :user, prefix: true
-
-  def user_is_verified
-    unless self.user_verified
-      errors.add(:verified, 'Only verified users may submit applications')
-    end
-  end
 end


app/models/scholarship_application.rb
```diff
@@ -1,17 +1,9 @@
 class ScholarshipApplication < ApplicationRecord
-  validate :user_is_verified
   validates :user_id, uniqueness: {scope: :scholarship_id}
 
   belongs_to :user
   belongs_to :scholarship
 
   delegate :name, to: :user, prefix: true
   delegate :email, to: :user, prefix: true
-  delegate :verified, to: :user, prefix: true
```
nellshamrell: This tells me there might be a method in the user model related to a user being verified - can you check whether there is one there that should be removed as well?

wimo7083: https://github.com/OperationCode/operationcode_backend/blob/5c3152374851db4a8607b73a0dd66e0e85571a98/app/models/user.rb#L9

```ruby
 LEADER = 'community leader'

  # Validation of military_status
  CURRENT = 'current'
  VETERAN = 'veteran'
  SPOUSE = 'spouse'
  MILITARY_STATUSES = [CURRENT, VETERAN, SPOUSE, nil]
  validates :military_status, inclusion: { in: MILITARY_STATUSES }
```

This is what I see for those. I wasn't sure to remove them as I don't know how these affect the current database.

wimo7083: Yup they ran locally and travis ran well too. 

```powershell
Finished in 24.010227s, 7.7467 runs/s, 25.4058 assertions/s.

186 runs, 610 assertions, 0 failures, 0 errors, 2 skips

You have skipped tests. Run with --verbose for details.
Inspecting 251 files
...........................................................................................................................................................................................................................................................

251 files inspected, no offenses detected
C:\Users\wimo7\Desktop\back [cancel_id_me +1 ~0 -0 !]>
```

PR #410, opened 2018-10-01
add Sidekiq web UI to admin section; go max-Sidekiq with jobs

### LINKED ISSUE #405: Sidekiq monitoring
# Add to dashboard sidekiq (job queue) monitoring
Without getting into the technical details, setting up kubectl, there's not an easy way to monitor sidekiq jobs. 

Our recent job queue issue happened because we had a billion retries, and we had no idea that jobs were backing up. This would give a data point to the health of the queue. 

One solution: https://github.com/mperham/sidekiq/wiki/Monitoring

Would allow us to put our monitoring behind a dashboard like this: 
image

User friendly monitoring for sidekiq.

robbkidd:
# Description of changes

This adds the Sidekiq web UI to the admin section under `/admin/sidekiq` for some better insight into the job queues. It also turns all the ActiveJob job classes into Sidekiq::Worker classes.

My recommendation is to commit to Sidekiq as the backend's background job processor. Some benefits of making backend's jobs full-fledged Sidekiq::Worker classes:

* more refined controls over when and how a job is executed
* easier test configuration to prevent the actual scheduling of jobs during a test run into a development environment
* increased visibility into the state of jobs through the Sidekiq administrative web UI

# Issue Resolved

Fixes #405

### DIFF
--- a/test/jobs/add_guest_to_send_grid_job_test.rb
+++ b/test/jobs/add_guest_to_send_grid_job_test.rb
@@ -1,12 +1,13 @@
 require 'test_helper'
 
-class AddGuestToSendGridJobTest < ActiveJob::TestCase
+class AddGuestToSendGridJobTest < ActiveSupport::TestCase
   test 'it adds the guest to send grid' do
+    job = AddGuestToSendGridJob.new
     guest = SendGridClient::Guest.user(valid_email)
 
     SendGridClient.any_instance.expects(:add_user).with(guest)
 
-    AddGuestToSendGridJob.perform_now(valid_email)
+    job.perform(valid_email)
   end
 end
 
--- a/app/controllers/api/v1/email_list_recipients_controller.rb
+++ b/app/controllers/api/v1/email_list_recipients_controller.rb
@@ -4,7 +4,7 @@ module Api
       def create
         raise "Invalid email address: #{permitted_params[:email]}" unless valid_email?
 
-        AddGuestToSendGridJob.perform_later(permitted_params[:email])
+        AddGuestToSendGridJob.perform_async(permitted_params[:email])
 
         render json: { email: permitted_params[:email], guest: true }, status: :created
       rescue StandardError => e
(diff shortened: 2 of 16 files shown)


robbkidd:
Not done yet is adding any sort of visible link _to_ the Sidekiq web UI. An admin would have to log in to the admin interface face and then type `.../admin/sidekiq` to get to it.

robbkidd:
I figured out adding a link to ActiveAdmin! RTFM: Sometimes It Helps™

This now adds a "Background Jobs" link to the default ActiveAdmin menu which will open `/admin/sidekiq` in a new tab/window (because there are not links back to the OC web site from the Sidekiq web UI).

screen_shot_2018-10-03_at_12_40_05_pm

test/jobs/add_guest_to_send_grid_job_test.rb
```diff
@@ -1,12 +1,13 @@
...
-class AddGuestToSendGridJobTest < ActiveJob::TestCase
+class AddGuestToSendGridJobTest < ActiveSupport::TestCase
   test 'it adds the guest to send grid' do
+    job = AddGuestToSendGridJob.new
     guest = SendGridClient::Guest.user(valid_email)
 
     SendGridClient.any_instance.expects(:add_user).with(guest)
 
-    AddGuestToSendGridJob.perform_now(valid_email)
+    job.perform(valid_email)
```
wimo7083: Since this job is derived from the sidekiq::Worker class we use the `perform` and `perform_async` methods?

Does `perform_now` and `perform_later` hook into the new sidekiq::Worker?

robbkidd: > Since this job is derived from the sidekiq::Worker class we use the perform and perform_async methods?

Correct!

> Does perform_now and perform_later hook into the new sidekiq::Worker?

No. Because these classes no longer descend from ActiveJob, they do not have now/later methods. That's the ActiveJob API. Sidekiq workers have `perform` (now) and `perform_async` (later).

PR #412, opened 2018-10-01
[Issue #376] Filter Expired Scholarships

### LINKED ISSUE #376: Filter expired scholarships from API response
# Feature

We're wanting to prevent display of expired scholarships on the Scholarships page.The API call to `/api/v1/scholarships` currently returns all scholarship objects, regardless of whether or not they are expired. To prevent the front-end from displaying the expired scholarships, this would currently require either:
a) Admins to manually remove expired scholarships from the DB, which is time-consuming, prone to human error/forgetfulness, and also would cause us to lose valuable data on scholarships
b) Filter out expired scholarships on the front-end and only show the ones that aren't expired
Each scholarship has a `close_time` property which represents the date when it's old/expired.  We can use that date to filter out those expired scholarships from API response, either when fetching from the existing `/api/v1/scholarships` endpoint (maybe adding some query params to signal that we only want current ones?), or from a new endpoint (something like `/api/v1/currentscholarships`). 
@hpjaj I  defer to you on implementation details.

colto:
# Description of changes
Create an 'Unexpired Scope' which filters out expired scholarships based on `close_time`. Then use this scope for the `scholarships#index`

# Issue Resolved

Fixes #376

### DIFF
--- a/app/models/scholarship.rb
+++ b/app/models/scholarship.rb
@@ -1,3 +1,5 @@
 class Scholarship < ApplicationRecord
   has_many :scholarship_applications
+
+  scope :unexpired, -> { where('close_time > ? OR close_time IS NULL', Time.current) }
 end
--- a/test/models/scholarship_test.rb
+++ b/test/models/scholarship_test.rb
@@ -1,0 +1,16 @@
+require 'test_helper'
+
+class ScholarshipTest < ActiveSupport::TestCase
+  test 'unexpired returns unexpired scholarships' do
+    nil_date = create :scholarship, close_time: nil
+    active_date = create :scholarship, close_time: 3.days.from_now
+
+    assert_equal Scholarship.unexpired, [nil_date, active_date]
+  end
+
+  test 'unexpired doesn\'t return expired scholarships' do
+    expired_date = create :scholarship, close_time: 3.days.ago
+
+    assert_not_includes Scholarship.unexpired, expired_date
+  end
+end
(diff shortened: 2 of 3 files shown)


app/models/scholarship.rb
```diff
@@ -1,3 +1,5 @@
 class Scholarship < ApplicationRecord
   has_many :scholarship_applications
+
+  scope :unexpired, -> { where('close_time > ? OR close_time IS NULL', Time.now) }
```
robbkidd: What do you think about `Time.current` here instead of `Time.now`? The ancient Rails maxims advise against `Time.now` because it is unaware of time zones.

From a business perspective, I'm not sure hour-precision is necessary for scholarship closing times, but I remember my experiences submitting things and they were usually last minute.

test/models/scholarship_test.rb
```diff
@@ -0,0 +1,16 @@
...
+    nil_date = create :scholarship, close_time: nil
+    active_date = create :scholarship, close_time: 3.days.from_now
+
+    assert_equal Scholarship.unexpired, [nil_date, active_date]
+  end
+
+  test 'unexpired doesn\'t return expired scholarships' do
+    expired_date = create :scholarship, close_time: 3.days.ago
+
+    assert_not_equal Scholarship.unexpired, [expired_date]
```
robbkidd: Assertions about "not equal" can come back to bite. An assertion that the results do not include the expired scholarship is a bit more robust.

```ruby
assert_not_includes Scholarship.unexpired, expired_date
```

colto:
@robbkidd I'm happy to rebase in that change once it's merged in. Otherwise, I can pluck out the VCR changes.

robbkidd:
@colto The changes look good and #413 merged to master branch, so this could be rebased for removing the cassettes.

wimo7083:
@robbkidd is this good to merge with the changes?

PR #413, opened 2018-10-02
Avoid bombarding geocode api during tests

robbkidd:
# Description of changes

Use geocoder's test mode and stubbing mechanism.

Instead of managing a vast collection of VCR cassettes of recorded API responses for geocode lookups, this turns on [Geocoder's test lookup](https://github.com/alexreisner/geocoder#testing). Otherwise, every test that builds or creates a user object will generate a real API request that VCR will either record and complain about or just complain about.

The user factory has a default zip of 97201 so that is added in the test_helper to Geocoder's stubbed locations with the real lat/log and state code for that zip. The vast majority of the tests that trigger geocoding lookup don't care about the actual geolocation data returned, so a default with weird data is also added to Geocoder. Some of the tests in `model/user_test` care about accurate lookup values, so those tests add stubbed locations specific to the test case. I opted to keep one cassette for the user test for non-US location lookup; that one uses the recorded "real" response from the API and then switches back to the test lookup provider.

Many cassettes died to bring us this information.

Also, set VCR to error by default for unknown 3rd party HTTP calls.

# Issue Resolved

No specific open Issue.


robbkidd:
Note, there are a lot of cassettes deleted in this PR, but only the ones that recorded the responses for GeoLoc lookups that are now unnecessary because of the stubbed responses.

PR #415, opened 2018-10-02
[Issue #397] Add provides_housing to Location

### LINKED ISSUE #397: Feature: Modify DB schema
# Collecting housing information for code schools.

We are currently collecting information on code school housing. We would like to be able to store this in our DB so that FE can render this content. 

Include with_housing in the location table, make sure this is send when a code school request occurs. And when performing migration set a default state of False. 
---

colto:
# Description of changes
Adding  `provides_housing` attr to Location Model

# Issue Resolved

Fixes #397

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180711212702) do
+ActiveRecord::Schema.define(version: 20181002181535) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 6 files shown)


colto:
@wimo7083 I'm a little confused. This is a location specific param, and it doesn't look like you are accepting nested params for locations in the code school (not sure that you can because code school has many locations). Do you want me to move this attr to the CodeSchool model?

robbkidd:
@colto I think @wimo7083 was repeating feedback from me and I was wrong! Looks like you've got it: `provides_housing` is a param on `Location`. 👍

db/schema.rb
```diff
@@ -124,8 +124,9 @@
     t.string   "state"
     t.integer  "zip"
     t.integer  "code_school_id"
-    t.datetime "created_at",     null: false
-    t.datetime "updated_at",     null: false
+    t.datetime "created_at",       null: false
+    t.datetime "updated_at",       null: false
+    t.boolean  "provides_housing"
```
robbkidd: Given the issue's statement:

> And when performing migration set a default state of False.

I would assume `false` is a reasonable default for any new record, though I agree it is not crystal clear.

colto: lol - I guess I just overlooked the basic requirements. Updating now.

test/factories/locations.rb
```diff
@@ -7,5 +7,6 @@
     city { Faker::Address.city }
     state { Faker::Address.state }
     zip { Faker::Address.zip_code }
+    provides_housing { [true, false].sample }
```
robbkidd: `Faker::Boolean.boolean` would have worked, too, but I find the simplicity of this refreshing. 👍

PR #416, opened 2018-10-04
Added meetup api member search and sync to update user data with meetup data

### LINKED ISSUE #87: Create job to merge Meetup user data into Postgres
# Feature
@rickr, we have a lot of data in Meetup that should be synced up with our regular user database, regarding members that go to our regular meetups, chapters and rate of growth, etc.

BRIMIL01:
# Description of changes
Added meetup api member search and sync to update user data with meetup data. I also added an editorconfig so my editor would "do the right thing" in regards to the rubocop settings. I wasn't able to actually make a call against the meetup api so my fixture data is best guess.

# Issue Resolved
Fixes #87

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,23 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+# Tab indentation (no size specified)
+[Makefile]
+indent_style = tab
+
+[**.rb]
+indent_style = space
+indent_size = 2
+
+# Matches the exact files either package.json or .travis.yml
+[.travis.yml]
+indent_style = space
+indent_size = 2
--- a/app/jobs/meetup_member_sync.rb
+++ b/app/jobs/meetup_member_sync.rb
@@ -1,0 +1,19 @@
+class MeetupMemberSync
+  include Sidekiq::Worker
+
+  def perform
+    meetup_client = Meetup.new
+    members = meetup_client.members_by_email
+
+    User.where(email: members.keys).each do |user|
+      member = members[user.email]
+      user.latitude ||= member['lat']
+      user.longitude ||= member['lon']
+      user.city ||= member['city']
+      user.state ||= member['state']
+
+      # Only update a user if values have changed
+      user.save if user.changed?
+    end
+  end
+end
--- a/test/support/meetups/sample_api_responses.rb
+++ b/test/support/meetups/sample_api_responses.rb
@@ -12,6 +12,12 @@ def event_endpoint_response
   response
 end
 
+def member_endpoint_response
+  response = build_response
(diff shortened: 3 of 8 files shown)


test/support/meetups/sample_api_responses.rb
```diff
@@ -36,32 +42,32 @@ def event_endpoint_parsed_response
...
-        'lat' => 36.844764709472656,
-        'lon' => -75.97899627685547,
-        'repinned' => false,
-        'address_1' => '1701 Baltic Avenue',
-        'city' => 'Virginia Beach',
-        'country' => 'us',
-        'localized_country_name' => 'USA',
-        'zip' => '',
-        'state' => 'VA'},
+        {'id' => 24628300,
```
BRIMIL01: My editor started a whitespace war :-(

.editorconfig
```diff
@@ -0,0 +1,23 @@
+# EditorConfig is awesome: https://EditorConfig.org
```
robbkidd: not gonna be part of your system!

(👍)

app/jobs/meetup_member_sync.rb
```diff
@@ -0,0 +1,19 @@
+class MeetupMemberSync
```
robbkidd: This job is sensible. I suggest a tweak to the naming, for it to end in `Job` to match the current naming pattern for the backend classes that are background jobs.

I'll offer `LocateUsersWithDataFromMeetupJob` unless there are thoughts about this worker being responsible other types of user data in the future.

BRIMIL01: Possibly `MeetupMemberToUserSyncJob`?

robbkidd: Maybe? That reads to me like we're creating users or something. `MeetupMemberSyncJob` could be fine if we want to keep the details of the sync out of the name.

wimo7083: `LoCaTe_MemBeRsMEeTUp4MeMBErs`

robbkidd: :shipit:

BRIMIL01: Names are hard

wimo7083: So this updates members if there is new meetup data?

`UpdateMembersMeetupDataJob`?

BRIMIL01: It doesn't overwrite, it just adds any missing data from a meetup call

robbkidd: I would describe the current behavior as "update a user with information about them from Meetup if they are a member of a Meetup." That information is currently location info nothing specific about Meetup, but could become more later.

BRIMIL01: Sure, maybe `ExternalUserAccountUpdateJob`

robbkidd: My vote is to add Job to the current class name and revisit the name when we want it to do more.

BRIMIL01: Done

PR #419, opened 2018-10-05
update native setup

wimo7083:
# Description of changes

Some more work on native setup instructions

### DIFF
--- a/docs/setup/native_setup.md
+++ b/docs/setup/native_setup.md
@@ -1,9 +1,59 @@
 # Native Setup
 
+### Quick Reference
+1. Install Dependencies (git, rails, redis, ...) 
+    
+    i. Install [git](#git) 
+
+    ii. Install [rails](#rails-setup)
+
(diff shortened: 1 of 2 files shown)


docs/setup/native_setup.md
```diff
@@ -28,13 +80,13 @@ git remote add upstream https://github.com/OperationCode/operationcode_backend.g
 
 #### Database Setup:
 
-
+1a. Run redis in separate tab.
 1. Run `cp config/database.yml. config/native_setup_database.yml`.
```
BRIMIL01: The current database.yml allows for passing a different host in via `POSTGRES_HOST`. I would recommend not having contributors replace the `database.yml` since it will likely be committed in PRs.

wimo7083: What would you recommend instead?

BRIMIL01: Passing the posgres host via an ENV variable. `POSTGRES_HOST=localhost bundle exec rails db:migrate`

PR #421, opened 2018-10-05
Test chatters

wimo7083:
# Description of changes

1. Reduce verbosity for tests
This reduces the amount of chatter seen while running unit tests. In cases where the rake file printed out debug information this adds an override so that the tests don't have this show. 

Not only is the easier to find unit test errors but stdout can be slow and the tests here run a little faster. 

```powershell

docker-compose run web bash -c 'export RAILS_ENV=test && rake db:test:prepare && rake test && rubocop'
Starting operationcodebackend_bundle_1             ... done
Starting operationcodebackend_operationcode-psql_1 ... done
Starting operationcodebackend_redis_1              ... done
Run options: --seed 61055

# Running:

......................................S.........................................................S..........................................................................................

Finished in 7.116397s, 26.2773 runs/s, 86.2796 assertions/s.
187 runs, 614 assertions, 0 failures, 0 errors, 2 skips

You have skipped tests. Run with --verbose for details.
Inspecting 251 files
...........................................................................................................................................................................................................................................................

251 files inspected, no offenses detected
```
2. Remove open and closed scope from the github statistic which can give warnings and confuse people.
3. Update gitignore to be okay with vscode
4. replace `before_filter` with `before_action` which was giving a deprecation warning.

PR #422, opened 2018-10-06
Test cleanup

BRIMIL01:
# Description of changes
Swaps out use of `assert` with `assert_equals` and fixed a timezone issue when running tests locally.

Based off #421

### DIFF
--- a/test/controllers/api/v1/airtable/mentorships_controller_test.rb
+++ b/test/controllers/api/v1/airtable/mentorships_controller_test.rb
@@ -10,7 +10,7 @@ class Api::V1::Airtable::MentorshipsControllerTest < ActionDispatch::Integration
     VCR.use_cassette('airtable/mentorship/successful') do
       get api_v1_airtable_mentorships_path, headers: @headers, as: :json
 
-      assert response.status == 200
+      assert_equal response.status, 200
       assert response.parsed_body.keys.sort == ['mentors', 'services', 'skillsets']
     end
   end
@@ -19,7 +19,7 @@ class Api::V1::Airtable::MentorshipsControllerTest < ActionDispatch::Integration
     VCR.use_cassette('airtable/mentorship/exceeded_rate_limit') do
       get api_v1_airtable_mentorships_path, headers: @headers, as: :json
 
-      assert response.status == 422
+      assert_equal response.status, 422
       assert response.parsed_body['error'].present?
     end
   end
@@ -42,7 +42,7 @@ class Api::V1::Airtable::MentorshipsControllerTest < ActionDispatch::Integration
         as: :json
       )
 
-      assert response.status == 201
+      assert_equal response.status, 201
       assert response.parsed_body['id'].present?
       assert response.parsed_body['createdTime'].present?
     end
@@ -66,7 +66,7 @@ class Api::V1::Airtable::MentorshipsControllerTest < ActionDispatch::Integration
         as: :json
       )
 
-      assert response.status == 201
+      assert_equal response.status, 201
       assert response.parsed_body['id'].present?
       assert response.parsed_body['createdTime'].present?
     end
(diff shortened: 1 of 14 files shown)


test/controllers/api/v1/airtable/mentorships_controller_test.rb
```diff
@@ -10,7 +10,7 @@ class Api::V1::Airtable::MentorshipsControllerTest < ActionDispatch::Integration
     VCR.use_cassette('airtable/mentorship/successful') do
       get api_v1_airtable_mentorships_path, headers: @headers, as: :json
 
-      assert response.status == 200
+      assert_equal response.status, 200
```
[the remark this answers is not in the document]
BRIMIL01: I _may_ have phoned it in a bit at the end

test/requests/airtable/mentorship_test.rb
```diff
@@ -54,25 +54,25 @@ def test_create_mentor_request_creates_the_passed_mentor_request
       response = Airtable::Mentorship.new.create_mentor_request(request_body)
 
       assert response['id'].present?
-      assert response.dig('fields', 'Slack User') == request_body[:slack_user]
-      assert response.dig('fields', 'Email') == request_body[:email]
-      assert response.dig('fields', 'Service') == [request_body[:services]]
-      assert response.dig('fields', 'Skillsets') == [request_body[:skillsets]]
-      assert response.dig('fields', 'Additional Details') == request_body[:additional_details]
-      assert response.dig('fields', 'Mentor Requested') == [request_body[:mentor_requested]]
+      assert_equal request_body[:slack_user], response.dig('fields', 'Slack User')
```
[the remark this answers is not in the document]
BRIMIL01: That is actually correct, the `request_body` is the hash of expected values.

test/requests/git_hub/client_test.rb
```diff
@@ -25,12 +25,12 @@ def test_successful_search_for_prs_for_multi_page
     ## link presence means there are multiple pages of results
     assert response.headers['link'].present?
     assert response['items'].present?
-    assert response.code.to_i == 200
-    assert pr['html_url'] == 'https://github.com/OperationCode/operationcode/pull/753'
-    assert pr['id'] == 240032714
-    assert pr['number'] == @pr_number
-    assert pr['title'] == 'Revert "waffle.io Badge"'
-    assert pr['state'] == 'closed'
+    assert_equal 200, response.code.to_i
```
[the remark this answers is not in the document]
BRIMIL01: That is correct

test/requests/git_hub/client_test.rb
```diff
@@ -116,28 +116,30 @@ def test_successful_commits_for
...
   def test_rate_limit_presence
     response = VCR.use_cassette('git_hub/commits/successful') do
       @client.commits_for(@oc, @pr_number, 1)
     end
 
-    assert response.headers['X-RateLimit-Remaining'] == '57'
-    assert response.headers['X-RateLimit-Limit'] == '60'
-    assert response.headers['X-RateLimit-Reset'] == '1503866476'
-    assert @client.send(:reset_time, response) == '8/27/2017 at 8:41:16pm UTC'
+    reset_time = Time.at(1503866476).strftime '%-m/%d/%Y at%l:%M:%S%P %Z'
```
wimo7083: How does this break test runs?

BRIMIL01: When running locally it uses the local time zone when generating the timestamp which doesn’t match the expected value which is UTC.

wimo7083: So I remember that sometimes the rate limiting test messes things up. We should probably have a way of testing rate limiting. 

Why is this done instead of some other way?

BRIMIL01: I can pull in something like timecop to manage the expected timezone

PR #431, opened 2018-10-16
319 slack integration 2

### LINKED ISSUE #319: Slack Integration 2: Add SlackUser Service Class
# Feature
Feature is being added to perform the actions required to manage modification of the `SlackUser` table.

Class will have two modification methods:
1. create
2. update

## Create Actions
1. Method uses provided email to query for `User`
2. User is found:
a. Creates a new `SlackUser` on FK `email`
b. Sends 201.
3. User is not found:
b. Creates a new `SlackUser` on FK `email` with no match of FK.
c. Sends 201
d. sends new `POST` to controlling endpoint for slack telling to send email update request.

## Exisiting User updates
1. Method uses provided email to query for `User`
2. User is found:
a. Updates `SlackUser: email`
b. Sends 201.
3. User is not found:
b. Creates a new `SlackUser` on FK `email` with no match of FK.
c. Sends 201
d. sends new `POST` to controlling endpoint for slack telling to send email update request.

Putting the `POST` in a separate action verify of reciept.
201 Response for both actions because the row is modified even if it has no FK match.

jdollar:
# Description of changes
Adds a create a update endpoint for slack integration issue #319 

# Issue Resolved
Fixes #319

### DIFF
--- a/app/controllers/api/v1/slack_users_controller.rb
+++ b/app/controllers/api/v1/slack_users_controller.rb
@@ -9,6 +9,7 @@ module Api
 
         ActiveRecord::Base.transaction do
           slack_user = SlackUser.new slack_user_params
+          slack_user.user_id = user.id unless user.nil?
           slack_user.save!
 
           render json: { action: generate_client_action(user, params[:slack_email]), slack_user: slack_user.id }, status: :created
--- a/app/controllers/api/v1/slack_users_controller.rb
+++ b/app/controllers/api/v1/slack_users_controller.rb
@@ -52,7 +52,7 @@ module Api
       end
 
       def generate_client_action(user, email)
-        user.nil? || user.email != email ? 'request_email_update' : 'none'
+        user.nil? || user.email != email ? 'request_account_verification' : 'none'
       end
 
       def slack_user_params
--- a/app/controllers/api/v1/slack_users_controller.rb
+++ b/app/controllers/api/v1/slack_users_controller.rb
@@ -2,9 +2,30 @@ module Api
   module V1
     class SlackUsersController < ApiController
       before_action :authenticate_py_bot!
+      before_action :set_slack_user, only: [:update]
 
       def create
-        # TODO: Placeholder for PyBot call to create new record
+        user = User.find_by(email: params[:slack_email])
+
+        ActiveRecord::Base.transaction do
+          slack_user = SlackUser.new slack_user_params
+          slack_user.save!
+
+          render json: { action: generate_client_action(user, params[:slack_email]), slack_user: slack_user.id }, status: :created
+        end
+      rescue StandardError => e
+        render json: { errors: e.message }, status: :unprocessable_entity
+      end
+
+      def update
+        user = @slack_user.nil? || @slack_user.user_id.nil? ? nil : User.find(@slack_user.user_id)
+        ActiveRecord::Base.transaction do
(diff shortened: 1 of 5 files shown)


app/controllers/api/v1/slack_users_controller.rb
```diff
@@ -28,6 +50,24 @@ def authenticate_py_bot!
       def auth_token
         @auth_token ||= request.headers.fetch('Authorization', '').split(' ').last
       end
+
+      def generate_client_action(user, email)
```
jdollar: @wimo7083 This is what I did for the 3d Items on the issue: "sends new POST to controlling endpoint for slack telling to send email update request."

Just return a action value that details if no further action needs to happen or if the client needs to request a email update to the user table

app/controllers/api/v1/slack_users_controller.rb
```diff
@@ -2,9 +2,31 @@ module Api
...
       before_action :authenticate_py_bot!
+      before_action :set_slack_user, only: [:update]
 
       def create
-        # TODO: Placeholder for PyBot call to create new record
+        user = User.find_by(email: params[:slack_email])
+
+        ActiveRecord::Base.transaction do
+          slack_user = SlackUser.new slack_user_params
+          slack_user.user_id = user.id unless user.nil?
```
jdollar: @wimo7083 On the slack_user table it had a user_id specified so I was populating that based on the user that returns from the e-mail address query.

Do have a concern with this that maybe the slack user has the e-mail specified for a normal user of the application then the slack_user will be related to a user that isn't theirs. I am assuming that the slack users are created through the py bot application exclusively? Maybe we don't want to populate the user_id right on create, but have a different action with pybot to be able to link a user and a slack_user? I'm not sure.

Just was a little concerned with setting the user ID and wanted to see if I should keep it or take it out?

app/models/slack_user.rb
```diff
@@ -1,3 +1,3 @@
 class SlackUser < ApplicationRecord
-  validates :slack_id, :slack_name, :slack_real_name, :slack_email, :user_id, presence: true
+  validates :slack_id, :slack_name, :slack_real_name, :slack_email, presence: true
```
jdollar: Removed the user_id from the validation since a slack_user has the possibility of being set without a user being defined in the main system

app/controllers/api/v1/slack_users_controller.rb
```diff
@@ -28,6 +50,24 @@ def authenticate_py_bot!
       def auth_token
         @auth_token ||= request.headers.fetch('Authorization', '').split(' ').last
       end
+
+      def generate_client_action(user, email)
+        user.nil? || user.email != email ? 'request_email_update' : 'none'
```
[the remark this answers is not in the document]
jdollar: Updated to be 'request_account_verification' vs 'request_email_update'

test/factories/slack_user.rb
```diff
@@ -0,0 +1,9 @@
+FactoryGirl.define do
+  factory :slack_user do
+    slack_id 'gobelygook'
```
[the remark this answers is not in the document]
jdollar: Updated to use the Faker library.

app/controllers/api/v1/slack_users_controller.rb
wimo7083: TODO: later- if we don't find a user_id matching slack log this somehow.

PR #434, opened 2018-10-19
Add a step to check for migrations and migrate DB if needed

### LINKED ISSUE #429: When merging to production perform db migrate
# Feature automatic db migration
We want to be able to perform db migrations without having to use a rails console and do things automatically.

1. Incorporate the migrate task into the prod build step. 
2. Include a new unit test that uses a default state prior to the migration and then performs a migration to the next (prefer only if a new migration file is in PR).

valeriecodes:
# Description of changes
Adds a step to the deploy process which would run migrations on the docker container before it is published.

**Note:** The original issue requested a test, but I'm not sure how to write an automated test for something like a Makefile task. I also am not sure how best to test locally, beyond running the new task in isolation (which I have done)

# Issue Resolved

Fixes #429

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,27 +3,27 @@ DOCKER := docker
 DOCKER_COMPOSE := docker-compose
 
 .PHONY: all
-all: run 
+all: run
 
 .PHONY:  nuke
-nuke: 
+nuke:
 	${DOCKER} system prune -a --volumes
 
-.PHONY: minty-fresh 
-minty-fresh: 
+.PHONY: minty-fresh
+minty-fresh:
 	${DOCKER_COMPOSE} down --rmi all --volumes
 
 .PHONY: rmi
-rmi: 
+rmi:
 	${DOCKER} images -q | xargs docker rmi -f
 
 .PHONY: rmdi
-rmdi: 
+rmdi:
 	${DOCKER} images -a --filter=dangling=true -q | xargs ${DOCKER} rmi
 
 .PHONY: rm-exited-containers
-rm-exited-containers: 
-	${DOCKER} ps -a -q -f status=exited | xargs ${DOCKER} rm -v 
+rm-exited-containers:
+	${DOCKER} ps -a -q -f status=exited | xargs ${DOCKER} rm -v
 
 .PHONY: fresh-restart
 fresh-restart: minty-fresh setup test run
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -95,7 +95,10 @@ bundle:
 
 setup: build db_create db_migrate
 
-publish: build
+migrate_if_needed:
+	bash -c 'if docker-compose run ${RAILS_CONTAINER} rake db:migrate:status | grep "^\s*down"; then docker-compose run ${RAILS_CONTAINER} rake db:migrate; fi'
```
robbkidd: I recently discovered the `db: abort_if_pending_migrations` task which pretty much does what it says: exits the rake execution with an error if there are pending migrations.

```
You have 10 pending migrations:
  20180129223616 AddGroupToTeamMembers
  20180204235116 AddMoocToCodeSchools
  20180401034441 CreateRoles
  20180408172532 AddRoleIdToAdminUser
  20180416024105 AddEmailToTeamMember
  20180420192231 AddIsPartnerToCodeSchools
  20180422030623 CreateSlackUser
  20180422031009 DropUserSlackId
  20180430161218 AddRepNameToCodeschool
  20180711212702 AddMilitarystatusToUser
Run `rails db:migrate` to update your database then try again.
```

And now that I know about it, of course I want to use it! What do you think about this?

```
migrate_if_needed:
	bash -c 'docker-compose run ${RAILS_CONTAINER} rake db:abort_if_pending_migrations || docker-compose run ${RAILS_CONTAINER} rake db:migrate'
```

Makefile
```diff
@@ -95,7 +95,10 @@ bundle:
...
+
+publish: build migrate_if_needed
 	bin/publish
```
robbkidd: 🤔 I'm not super familiar with the deployed production environment. Looking at the `bin/publish` script has me thinking that this wiring up of a `migrate_if_needed` action prior to publish would have the migrations occur in an environment that is _not_ production, but rather in a development or CI environment that is doing Docker image tagging and kubernetes service poking.

valeriecodes: Gotcha. I thought maybe the docker container that build was run on would then be promoted to production.

wimo7083:
https://github.com/OperationCode/operationcode_backend/wiki/Performing-Production-DB-Migration

Here's a wiki article I created now that our former lead has explained our previous steps, you can look for the details but it's as simple as:

1. connect to kubectl
2. get rails console 
3. `rake db:migrate`

PR #438, opened 2018-10-28
increase expectations

wimo7083:
# Increase time limit for emails

Since these are done manually we want to set more realistic expectations.


wimo7083:
Ashley has given me the thumbs up.

PR #443, opened 2018-11-04
Feature/local env documentation

igor-starostenko:
# Description of changes
- Fixed `config/database.yml` to allow to run locally without docker
- Fixed some links in documentation, did some cleanup and updated native setup docs
- Added vim files to `.gitignore`


README.md
```diff
@@ -62,9 +63,6 @@ Recommended versions of tools used within the repo:
 
 </details >
 
-## Description
-If you'd like to learn more about what this project is what a Backend is you can read about this in our [Introduction Guide](INTRODUCTION.md).
```
drewish: I guess this guide was removed?

config/database.yml
```diff
@@ -20,8 +20,8 @@ default: &default
   # For details on connection pooling, see rails configuration guide
   # http://guides.rubyonrails.org/configuring.html#database-pooling
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
-  user: postgres
-  host: <%= ENV['POSTGRES_HOST'].present? ? ENV['POSTGRES_HOST'] : 'operationcode-psql' %>
+  user: <%= ENV['POSTGRES_USER'] %>
```
drewish: It seems like you'd want something like `ENV['POSTGRES_USER'].presence || 'postgres'` to keep the current behavior.

config/database.yml
```diff
@@ -20,8 +20,8 @@ default: &default
...
+  host: <%= ENV['POSTGRES_HOST'] %>
```
drewish: What's the thought behind removing the default value?

docs/setup/native_setup.md
```diff
@@ -14,34 +16,50 @@ You are now ready for the actual **OperationCode Backend** code base.
...
+1. Run `bin/rails db:create`.
+2. Run `bin/rails db:migrate`.
+3. Run `bin/rails db:seed`.
 
 ### Testing
-1. Run `RAILS_ENV=test rake db:test:prepare`.
-2. Run `bin/rake test & rubocop`.
+
+1. Run `RAILS_ENV=test bin/rails db:test:prepare`.
+2. Run `bin/rails test`
```
drewish: Nit picky but the other ones end with periods and this doesn't... personally I'd just drop all the periods since they make it harder to quickly copy paste the commands but it would be nice to keep them all consistent one way or the other.

config/database.yml
igor-starostenko: The current behavior is kept by the `.env` file which is applied when running `docker-compose`

config/database.yml
igor-starostenko: Default values are the names of docker containers which should be applied only when you want to run the environment in docker. Thus they should be optional and we can use the `.env` file to read the defaults when we use docker.

docs/setup/native_setup.md
igor-starostenko: Yeah, I agree. Will remove the periods. Thanks

kylemh:
Thanks @igor-starostenko  ❤️

igor-starostenko:
@wimo7083, I have consolidated the docs from pull #419 
My main concern is that there are so many places in the docs where the same or similar instructions are listed. For example: there are at least 3 different places with the same git instructions. That's out of scope for this PR, but would be nice to better organize it going forward.

wimo7083:
That's good enough for me. I'll probably just close the other one if this one is sufficient to run locally. 
Thanks for the big lift.

.gitignore
```diff
@@ -27,6 +27,14 @@
 # VSCode
 .vscode/*
 
+# Vim
```
wimo7083: Found the vim user 💯

PR #455, opened 2018-11-09
remove secret_key, in place of environment variable

### LINKED ISSUE #441: Devise secret key is in source control
I was flipping through the source and noticed that we have the devise secret key committed to source control. This should be extracted as an environment variable.

https://github.com/OperationCode/operationcode_backend/blob/5df843a43619738df7900969b59c6edb009966bd/config/initializers/devise.rb#L9

wimo7083:
# Description of changes

Devise config currently has `secret_key` set in the config, this overrides `secret_key_base` which is set in the environment variables. This removes the `secret_key` from the configs for clarity. 

# Issue Resolved

Fixes #441

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -1,16 +1,15 @@
 # Use this hook to configure devise mailer, warden hooks and so forth.
 # Many of these configuration options can be set straight in your model.
 Devise.setup do |config|
-  
   # The secret key used by Devise. Devise uses this key to generate
   # random tokens. Changing this key will render invalid all existing
   # confirmation, reset password and unlock tokens in the database.
-  # Operation Code Comment: 
+  # Operation Code Comment:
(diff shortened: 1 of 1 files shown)


config/initializers/devise.rb
```diff
@@ -6,7 +6,6 @@
   # confirmation, reset password and unlock tokens in the database.
   # Devise will use the `secret_key_base` as its `secret_key`
   # by default. You can change it below and use your own secret key.
-   config.secret_key = '34486348de8caa31dbf426f515084b3c4b33c15954ce1b0eb8f664f58eb68453ba7dd2603aa0a6a54ee816adcb5a72adca4564df5d772a7ae45ab1efbf814c9c'
```
kylemh: Can you change the comments above to reflect the correct way of changing keys?

wimo7083:
Pod has been re-deployed, verified this is working by performing a password reset, which worked.

PR #459, opened 2018-11-10
add user and password values to prod

wimo7083:
# Description of changes

# Issue Resolved

Fixes #

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -35,4 +35,6 @@ test:
 production:
   <<: *default
   host: <%= ENV['POSTGRES_HOST'].present? ? ENV['POSTGRES_HOST'] : 'operationcode-psql-postgresql' %>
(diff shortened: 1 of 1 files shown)


config/database.yml
```diff
@@ -35,4 +35,6 @@ test:
 production:
   <<: *default
   host: <%= ENV['POSTGRES_HOST'].present? ? ENV['POSTGRES_HOST'] : 'operationcode-psql-postgresql' %>
+  user: <%= ENV['POSTGRES_USER'] %>
```
kylemh: I believe `<<: *default` is like a spread operator.

I think you can leave these out and only define `database`

wimo7083: Yeah I was still getting missing env var issues. But that was because the redis pod isn't updating when this deploys. I updated the pod and it worked. Going to revert this change, and see if it breaks.

PR #468, opened 2018-12-29
updates code school url when protocol is missing

NeimadTL:
tries with https first if possible then http.

### DIFF
--- a/app/models/code_school.rb
+++ b/app/models/code_school.rb
@@ -2,4 +2,23 @@ class CodeSchool < ApplicationRecord
   validates :name, :url, :logo, presence: true
   validates_inclusion_of :full_time, :hardware_included, :has_online, :online_only, :in => [true, false]
   has_many :locations, -> { order('state ASC, city ASC') }, dependent: :destroy
+
+  before_create :check_scheme
+
+  private
+
+  def check_scheme
+    uri = URI.parse(url)
+    update_url(uri) if uri.scheme.nil?
+  end
+
+  def update_url(uri)
+    candidate = 'https://' << uri.to_s
+    begin
+      HTTParty.get(candidate, timeout: 2)
+    rescue StandardError
+      candidate.sub!('s', '')
+    end
+    self.url = candidate
+  end
 end
--- a/test/models/code_schools_test.rb
+++ b/test/models/code_schools_test.rb
@@ -16,4 +16,26 @@ class CodeSchoolTest < ActiveSupport::TestCase
   test 'rep_email should be present' do
     assert @code_school.rep_email.present?
   end
+
+  test 'url should be present' do
+    assert @code_school.url.present?
+  end
+
+  test 'url does not change when url is correct' do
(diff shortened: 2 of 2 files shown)


app/models/code_school.rb
```diff
@@ -2,4 +2,23 @@ class CodeSchool < ApplicationRecord
...
   has_many :locations, -> { order('state ASC, city ASC') }, dependent: :destroy
+
+  before_create :check_scheme
+
+  private
+
+  def check_scheme
+    uri = URI.parse(url)
+    update_url(uri) if uri.scheme.nil?
+  end
```
seanlinsley: Currently `check_scheme` reads the URL attribute, then `update_url` writes it.

To make the URL handling code reusable the API might look something like this:
```ruby
self.url = ensure_url_protocol(url)
```
For now though, I'd probably keep it all in one method.

app/models/code_school.rb
```diff
@@ -2,4 +2,23 @@ class CodeSchool < ApplicationRecord
...
+
+  def update_url(uri)
+    candidate = 'https://' << uri.to_s
+    begin
+      HTTParty.get(candidate, timeout: 2)
+    rescue StandardError
+      candidate.sub!('s', '')
+    end
+    self.url = candidate
```
seanlinsley: There are two bugs that I see:

1. a valid URL to a website that does not support HTTPS would have an "s" characters removed
2. a invalid URL that results in a request failure would be saved anyway (again with an "s" character removed)

The first issue can be resolved by using the Addressable gem:
```ruby
def check_scheme
  if url !~ /\Ahttps?:\/\//
    uri = Addressable::URI.heuristic_parse(url, scheme: 'https')

    begin
      HTTParty.get(uri.to_s, timeout: 2)
    rescue # (no need for StandardError. that's the default)
      uri.scheme = 'http'
    end

    self.url = uri.to_s
  end
end
```
It'd be great to use the standard library instead of a gem, but Addressable works around this unfortunate behavior of URI:
```ruby
uri = URI('foo.com')
uri.scheme = 'https'
uri.to_s # https:foo.com
```

For the second issue, we can update the code to `retry` the request after updating the protocol, to ensure it's really valid. Then if it still doesn't work we can add a validation error to prevent the record from saving:
```ruby
def check_scheme
  uri = Addressable::URI.heuristic_parse(url, scheme: 'https')

  begin
    HTTParty.get(uri.to_s, timeout: 2)
  rescue
    if uri.scheme == 'https'
      uri.scheme = 'http'
      retry
    else
      errors.add :url, 'unable to verify URL'
      return
    end
  end

  self.url = uri.to_s
end
```

But I'm new here, so I'm not sure if y'all want to show a validation error here or not.

seanlinsley: I believe that HTTParty only raises exceptions if it isn't able to connect to the server. If it receives a 404, though, you'd have to check the response object. But that might not be something y'all are trying to protect against.

PR #474, opened 2019-03-21
add slack username validation

Brantron:
# Description of changes
This should validate the slack username if this client is still working. Can't test locally because I don't have the creds

### DIFF
--- a/app/lib/airtable/mentorship.rb
+++ b/app/lib/airtable/mentorship.rb
@@ -36,6 +36,8 @@ module Airtable
         }
       }.to_json
 
+      raise Airtable::Error("#{body[:slack_user]} is not a valid slack username") unless Slack::Utils.new.username_is_registered?(body[:slack_user])
+
       client.post_record('Mentor Request', request_body)
     end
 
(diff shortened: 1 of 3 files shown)


app/lib/airtable/mentorship.rb
```diff
@@ -52,6 +52,13 @@ def data_for(table, name_key: 'Name')
       end
     end
 
+    def validate_user(email)
+      return if Rails.env.test?
```
wimo7083: Will eventually want to fix the stubbing so that we can test the negative case.

app/lib/slack/utils.rb
```diff
@@ -4,11 +4,8 @@ module Slack
   # Utilities class for running methods and discovery on Client
   class Utils
     def email_is_registered?(email)
-      users_list = client.fetch_users_list
-      users_list['members'].each do |m|
-        return true if m['profile']['email'] == email
-      end
-      false
+      user_response = client.lookupByEmail(email)
```
wimo7083: I'm pretty certain this is going to not work. Because the client doesn't have a method for this.

PR #484, opened 2019-04-19
Upgrade container image base to ruby2.4

ohaiwalt:
# Description of changes

Testing to see if upgrading ruby still works 🙏 
# Issue Resolved

Fixes #


ohaiwalt:
There was an attempt! This will require further investigation, however.

PR #493, opened 2019-04-20
Http thing

wimo7083:
# Description of changes

# Issue Resolved

Fixes Security vuln related to cookies.


kylemh:
We'll need [to set cookies from the relevant responses](https://api.rubyonrails.org/classes/ActionDispatch/Cookies.html) as well.

PR #494, opened 2019-04-21
Use cookies in auth

kylemh:
# Description of changes

- Attempting to set cookies on responses related to authentication


config/environments/production.rb
```diff
@@ -32,7 +32,7 @@
   # config.action_cable.allowed_request_origins = [ 'http://example.com', /http:\/\/example.*/ ]
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  # config.force_ssl = true
+  config.force_ssl = true
```
wimo7083: Will you add a comment to staging.rb stating why we don't do ssl in staging with the difficulties of local development.

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -9,13 +9,26 @@ def create
         self.resource = warden.authenticate!(auth_options)
         sign_in(resource_name, resource)
 
-        set_sso_response if sso_request?
-        @redirect_path ||= '/profile'
+        cookies[:token] = resource.token
```
wimo7083: Later we should do something in the user model or sessions model for signins where we set the token.

wimo7083: This didn't make much sense to me. Here's what I'm suggesting. anywhere we would normally call `sign_in` we now call some new thing `sign_in_cookies` which does the cookies part for us everywhere.

kylemh: That seems fair. How do I alter the `sign_in` method in that regard

jaydorsey: You could stick a small helper method in your ApiController that takes a user & additional optional args, if they all inherit from that.

I only see the `sign_in` method used 3x in the codebase, and never in the same context (different numbers of args). Anything that you add is probably going to have at least as many lines of code, and maybe even more. I don't think having it inline in the controller(s) is super horrible

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -9,13 +9,26 @@ def create
...
+
+        @redirect_path_login ||= '/profile'
```
jaydorsey: You could probably remove this line, in favor of using a URL helper inline w/ the redirect_to. This _shouldn't_ really need to be memoized, unless there's another way to set it that I'm missing (e.g. some kind of before hook that _might_ set it)

If that's the case, then this needs to stay

app/controllers/api/v1/sessions_controller.rb
```diff
@@ -9,13 +9,26 @@ def create
...
+          redirect_to: @redirect_path_login
+        }
+      end
+
+      def destroy
+        sign_out @user
+
+        cookies.delete :token
+
+        @redirect_path_logout ||= '/login'
```
jaydorsey: Same thing down here, unless it possibly gets set somewhere else

PR #497, opened 2019-04-21
Add bundle prefixes to rails/rake/rubocop commands

jaydorsey:
# Description of changes

We're isolated in the docker containers so we _shouldn't_ run into conflicts here, but this closely reflects the commands as you'd want to run them locally, if you weren't using a Makefile. So, I'd consider it more a "documentation" fix than anything

Also cleaned up some trailing empty spaces on commands. 

This might be considered a style/preference but I consider it more of a good rule to follow when using bundler. Please feel free to close it if it doesn't add any value.

# Issue Resolved

No issues closed. This is a general documentation/cleanup suggestion

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,34 +3,34 @@ DOCKER := docker
 DOCKER_COMPOSE := docker-compose
 
 .PHONY: all
-all: run 
+all: run
 
 .PHONY:  nuke
-nuke: 
+nuke:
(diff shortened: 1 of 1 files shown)


jaydorsey:
As an FYI, I actually just ran into this issue on a separate PR that I'm working on, so I guess it's not _quite_ as much of "just" a style issue, or isolated because it's in docker, as I thought

image

PR #504, opened 2019-04-21
WIP: API request to PyBot for Slack invite

### LINKED ISSUE #481: ActiveJob to send a new user notification to the PyBot
# Feature

Current ActiveJob sends new user event directly to Slack API. This logic instead can be handled entirely with PyBot.

Create ActiveJob to forward user events to PyBot. With user email and the user details like interests, mentor etc.

This task depends on ticket in PyBot to complete https://github.com/OperationCode/operationcode-pybot/issues/67

juliantrueflynn:
# Description of changes
Have PyBot handle Slack API invite. This is apart of ongoing task of moving responsibility to PyBot instead of Rails.

Note: this is before adding tests to make sure is on right track.

# Issue Resolved

Fixes #481


app/jobs/slack_jobs/inviter_job.rb
```diff
@@ -3,14 +3,10 @@ class InviterJob < SlackJobs
...
-      # Invites user to general channel
       Raven.user_context email: email
       Raven.tags_context slack_invite: 'yes'
-      slack_client.invite(
-        email: email,
-        channels: ['C1T66M57G'],
-        extra_message: 'Welcome to Operation Code! After logging in, please introduce yourself in the #general channel - a bit about who you are, where you\'re coming from, and where you\'re interested in going next.'
-      )
+
+      PybotSlackClient.new.invite(email)
```
wimo7083: Since this is happening in the background with a sidekiq worker I need to make sure the sidekiq/ redis container has access to the environment variables.

app/lib/pybot_slack_client.rb
```diff
@@ -0,0 +1,32 @@
...
+    response = self.class.get(
+      '/verify',
+      headers: headers,
+      query: { email: email }
+    )
+
+    response.parsed_response
+  end
+
+  def invite(email)
```
wimo7083: This has two states:

No Need to retry
Need to retry

Some of the responses I've seen can be:

No Exception - explicitly allow these get with @AllenAnthes to figure out how he's going to return:
User success invite - 200 from service
User failed invite - user already invited  - 
User failed invite - bad user email thing+mail@gmail.com is an example 

Raise exception For any thing else
User failed invite - error connecting to service - timeout or something
User failed invite - bad service token - 403
User failed invite - service used an invalid token to talk to slack - whatever service sends us

app/lib/pybot_slack_client.rb
```diff
@@ -0,0 +1,32 @@
+class PybotSlackClient
+  include HTTParty
+
+  base_uri ENV['SLACK_SERVICE_URL']
+
+  attr_reader :headers
+
+  def initialize
```
wimo7083: In general we should put logging around these requests too. I want to be able to open up the logs and see what's happening for when we have failures. Something like sending request, and logging the request response in addition to everything else should be good.

app/lib/pybot_slack_client.rb
wimo7083: After discussion this with @AllenAnthes  raise an exception unless it's a 200 since he's handling all the good or bad things.

app/jobs/slack_jobs/inviter_job.rb
```diff
@@ -3,14 +3,10 @@ class InviterJob < SlackJobs
     include Sidekiq::Worker
 
     def perform(email)
```
wimo7083: Also let's set the max retries like this: 
https://github.com/mperham/sidekiq/wiki/Advanced-Options#workers

details on wait time are here:
https://github.com/mperham/sidekiq/wiki/Error-Handling#automatic-job-retry

I think setting it to 5 retries gets at about 10 minutes from initial try to unique death handler in which we need to notify the mods to read out.

app/lib/pybot_slack_client.rb
juliantrueflynn: Do you want it to raise an error if the `response.status != 200` and or if `response.ok === 'fail'` in response?

wimo7083: Per allen:
```json
{
    "ok": false,
    "error": "already_in_team"
}
```
returns a 200 from pybot service. For now let's just throw an error if status != 200, then we log them all and we can add cases later.

app/jobs/slack_jobs/inviter_job.rb
juliantrueflynn: I just tested this with 5 retries set, in case you were curious here is the results:
2019-04-23T05:14:28.630184
2019-04-23T05:14:56.946850
2019-04-23T05:16:03.605952
2019-04-23T05:16:54.764033
2019-04-23T05:19:12.275025

wimo7083:
Theres some randomness involved. Let's go for somewhere between 5 to 10. I'll trust your judgement

juliantrueflynn:
> Theres some randomness involved. Let's go for somewhere between 5 to 10. I'll trust your judgement

The PR I just made has `retries: 5`, do you want to go above that? You're right it's random and the test I did was slightly under 5 minutes.

juliantrueflynn:
Sorry, I swore I pushed the retries ActiveJob update but I guess I didn't? Hope you weren't waiting for this.

app/jobs/slack_jobs/inviter_job.rb
```diff
@@ -1,16 +1,13 @@
 class SlackJobs
   class InviterJob < SlackJobs
     include Sidekiq::Worker
+    sidekiq_options retry: 5
```
wimo7083: I like this.

.env
```diff
@@ -3,3 +3,5 @@ REDIS_URL=redis://redis:6379/0
 RAILS_ENV=development
 POSTGRES_USER=postgres
 POSTGRES_HOST=operationcode-psql
+SLACK_SERVICE_URL=http://pybot.operationcode.org:8080/pybot/api/v1/slack
```
wimo7083: This is the url for the pybot / slack service url endpoint for slack actions.

app/jobs/slack_jobs/inviter_job.rb
```diff
@@ -1,16 +1,13 @@
...
-      # Invites user to general channel
       Raven.user_context email: email
       Raven.tags_context slack_invite: 'yes'
-      slack_client.invite(
-        email: email,
-        channels: ['C1T66M57G'],
-        extra_message: 'Welcome to Operation Code! After logging in, please introduce yourself in the #general channel - a bit about who you are, where you\'re coming from, and where you\'re interested in going next.'
-      )
+
+      SlackService::Client.new.invite(email)
```
wimo7083: In a later PR, we're going to add tests surrounding creating the job.

